PR #2, opened 2016-04-20
Initial rename to geoviews and tidy.

bjlittle:


philippjfr:
Looks good thanks, there's lots of bits floating that needs renaming/fixing so I'm happy to do that in another PR.

PR #5, opened 2016-04-20
Minor fixes for the Introductory Tutorial

philippjfr:
Small changes for the Introductory Tutorial, demonstrating select instead of creating a DynamicMap to sample. Additionally removed coastline scale as '1000m' is not a valid scale and the correct way to set the scale is via a plot option.


bjlittle:
:+1:

PR #10, opened 2016-04-20
Updated references to holocube in docs

philippjfr:
Small PR to replace references to holocube in docs, using geoviews instead. Also includes small fix for Homepage example notebook.

### DIFF
--- a/doc/Homepage.ipynb
+++ b/doc/Homepage.ipynb
@@ -81,7 +81,7 @@
    },
    "outputs": [],
    "source": [
-    "%%opts GeoImage [colorbar=True] (cmap='viridis')\n",
+    "%%opts Image [colorbar=True] (cmap='viridis') Overlay [fig_size=200]\n",
     "(hv.Dataset(surface_temp).groupby(['time'], group_type=gv.Image) * gv.Feature(cf.COASTLINE))"
    ]
   }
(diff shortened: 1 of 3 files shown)


jbednar:
How does the options system know to apply the given options to gv.Image and not hv.Image?

philippjfr:
It doesn't, it will apply it to all Image types displayed in that cell. We discussed this today and decided that's okay behavior where the semantics of the two types match exactly and they share all plot/style options. Currently Image, Points and Text have such overlap and we've decided that we should split the Contours type into LineContours and FilledContours, because those types are not equivalent to the HoloViews Contours type.

jbednar:
Ooh, that's tricky -- semantically, we used to be able to say that we apply the options by Python type, but now it's clear that we're instead applying them by the class name, as a textual string.  So we now need to ensure that all classes that inherit from hv.Element are semantically equivalent if they have the same class name, regardless of which module they appear in.  

So we now need to state that very clearly in the docs about extending HoloViews -- don't make a new class have the same name as an existing class, unless you want the existing class's options to apply to the new class too. Even that's probably ok; people can handle that.

But then the bad thing will be when a user has previously written their own class that derives from hv.Element, and we then create a class with the same name, probably because both of us have realized that a plot of that type was missing before.  At that point the user code will break (i.e., have incorrect or incompatible options applied), which is ordinarily something that Python namespaces help people avoid but which we are now exposed to because we're not respecting namespaces.

PR #13, opened 2016-04-21
Name changed to Met Office, not UK Met Office

corinnebosley:


bjlittle:
Are we really sure about this ? :smile:

bjlittle:
No, really ... are we ?

corinnebosley:
I don't know, I'm just trying to do what I'm told.  Richard says be the brand, I'll be the brand...

bjlittle:
@corinnebosley Awesome, thanks!

jbednar:
For what it's worth, a google search on "Met Office" only pulls up pages about the UK's Met Office until the middle of the second page, where the Trinidad and Tobago Met Office timidly raises its hand.  So I think the brand is solid enough not to need "UK".  Making people know that it originally stood for Meteorological Office, which makes so much more sense than Metropolitan, is another matter. :-)

rhattersley:
> No, really ... are we ?

If you want to get _really_ picky :wink: I think "they" prefer to treat it as a name and avoid even using the definite article: so "the Met Office" becomes just "Met Office" (just like we have "Continuum Analytics" and not "the Continuum Analytics")... but I'm not going to push for it.

corinnebosley:
I'm gonna change it to 'those weather people'...

jbednar:
We *are* allowed to blame you guys for the weather, right? :-)

jbednar:
> "the Met Office" becomes just "Met Office" 

Hmm; so it can be "the Met" or "Met Office" but not "the Met Office"?  No problem!

rhattersley:
This might be getting out of hand! I've seen lots of uses of "_the_ Met Office" even in official material (e.g. the [home page](http://www.metoffice.gov.uk/) itself!), so let's not go changing that! As far as I'm concerned this PR is perfect as it is - I was just joking earlier!

Now... where's that "Lock conversation" button...? :wink:

rhattersley:
> We are allowed to blame you guys for the weather, right? :-)

Yes. :grin:

rhattersley:
> Hmm; so it can be "the Met"

Definitely not that one! "The Met" is short for London's [Metropolitan Police Service](http://content.met.police.uk/Site/About). They're a very different bunch! :wink:

PR #16, opened 2016-04-22
Homepage fixes

jbednar:
The homepage had gotten out of date now that cube support has moved directly into HoloViews.  I also simplified the example to omit the cube summary, because keeping it as bare-bones as possible on the home page seems crucial, so that people don't get distracted by studying the summary output.  That last change is in a separate commit in case someone wants to roll it back; as you wish!  I also added a basic logo and favicon, which can of course be updated to any suitable image anyone prefers.


jbednar:
I also edited the intro tutorial.  I found the comments interspersed with the code to be distracting -- the tendency is to try to read the comments, when the code itself is already fairly readable, and we'd much rather people look directly at the code to figure out what's going on, than to first have to parse a comment and then consider looking at the code.  And the comments weren't always correct; e.g. it said it was looking at 0,10, when it was really looking at 0,0. Fixed it to look at 0,10, since that's where the point was located.  Also changed the various obj.to.*type*( calls to obj.to(gv.*Type*), which I think is vastly more explicit and clearer, and not really any longer as text.

philippjfr:
I'd like to merge this asap, as it has changes to the Introductory Tutorial, which should be merged before #17, so I can make some fixes to how the aspects are handled there.

Issue #15, opened 2016-04-22
Using Image and Line/FilledContours for non-geographic plots


philippjfr:
Until we have fully integrated the Image in HoloViews with the data interfaces the existing geoviews Image and Contours plots should also support plotting non-geographic data. Currently this works but has two major issues:

1) The iris plotting calls do not respect the order of dimensions resulting in mislabeling of the axes.

2) The aspect on these plot types is set to 'equal' by default which is sensible for 'longitude' and 'latitude' since those exist in the same space but often results in very wide or tall plots.

The aspect issue should be easy enough to fix as we can just detect whether a plot is geographic using the ``is_geographic`` utility function in ``geoviews.element.geo`` and set the aspect accordingly.

The axis mislabeling is more problematic as it will require us to make changes to iris plotting functions so we can respect the order of coordinates in the cube if desired.

philippjfr:
This is now supported, closing.

PR #17, opened 2016-04-23
Improved support for geographic plot aspects and extents

philippjfr:
This PR makes three main improvements:

1. Fix for the ``is_geographic`` function to correctly handle Tiles, WMTS and Feature Elements.

2. Only set the aspect to equal on geographic plot types, avoiding the issue raised in #15.

3. Add a get_extents method, which makes use of ``GeoAxes.set_extent`` to correctly compute the plot extents based on the range and extents of the Element. This allows setting extents on the Element in its native coordinate system.

Letting cartopy handle computing the extent transform is good, but it would be nice if we could factor out the code in cartopy's ``GeoAxes.set_extents`` to compute the extent transform without having to set it.

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -10,24 +10,27 @@ from iris.cube import Cube
 
 geographic_types = (cGoogleTiles, cFeature)
 
-def is_geographic(dataset, kdims=None):
+def is_geographic(element, kdims=None):
     """
-    Small utility that determines whether the supplied dataset
-    and kdims represent a geographic coordinate system.
+    Utility to determine whether the supplied element optionally
+    a subset of its key dimensions represent a geographic coordinate
+    system.
     """
     if kdims:
-        kdims = [dataset.get_dimension(d) for d in kdims]
+        kdims = [element.get_dimension(d) for d in kdims]
     else:
-        kdims = dataset.kdims
+        kdims = element.kdims
 
-    if isinstance(dataset.data, geographic_types) or isinstance(dataset, WMTS):
+    if len(kdims) != 2:
+        return False
+    if isinstance(element.data, iris.cube.Cube):
+        return all(element.data.coord(kd.name).coord_system for kd in kdims)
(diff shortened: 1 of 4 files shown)


jbednar:
I've merged the other two pull requests that affect the Introductory Tutorial, so this one can be fixed up now.

philippjfr:
I've updated the PR and it should now be ready to merge, unless any issues are raised during review.

geoviews/element/geo.py
```diff
@@ -8,6 +8,27 @@
...
+    """
+    if kdims:
+        kdims = [dataset.get_dimension(d) for d in kdims]
+    else:
+        kdims = dataset.kdims
+
+    if isinstance(dataset.data, geographic_types) or isinstance(dataset, WMTS):
+        return True
+    if (len(kdims) == 2 and
+        ((isinstance(dataset, _Element) and kdims == dataset.kdims) or
```
philippjfr: This is missing ``and  dataset.crs``.

geoviews/element/geo.py
```diff
@@ -8,6 +8,27 @@
...
+        (isinstance(dataset.data, iris.cube.Cube) and all(dataset.data.coord(
```
jlstevens: So one of the conditions is that if the dataset is an ``_Element``, it is geographic if the list of kdims matches the dataset kdims. Won't this latter condition always be true if the kdims argument isn't supplied (you have ``else: kdims = dataset.kdims`` above).

jlstevens: I think it might be a little cleaner to pull out some of the isinstance checks as variables:

```python
is_cube = isinstance(element.data, iris.cube.Cube)
is_geoelement = isinstance(element, _Element)
```

jlstevens:
The conditional logic in ``is_geographic`` is currently a bit hairy but otherwise this looks fine and ready to merge.

philippjfr:
I've refactored the logic in ``is_geographic`` and correctly declared the dimensionality of some of the other classes.

jlstevens:
Looks good. I understand it is now ready to merge.

PR #18, opened 2016-04-23
Added Tutorial to explain how explore Iris Cubes with GeoViews

philippjfr:
The Tutorial uses live and dynamic modes so won't work properly on the website. It outlines how the conversion interface works.

I've already uploaded a build [to the website](http://geo.holoviews.org/Working_with_Iris_Cubes.html). It depends on improved axis labels for grids (https://github.com/ioam/holoviews/pull/636) and correct handling of aspects in https://github.com/ioam/geoviews/pull/17.

### DIFF
--- a/doc/Working_with_Iris_Cubes.ipynb
+++ b/doc/Working_with_Iris_Cubes.ipynb
@@ -1,0 +1,383 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "import iris\n",
+    "import numpy as np\n",
+    "import holoviews as hv\n",
+    "import geoviews as gv\n",
+    "from cartopy import crs as ccrs\n",
+    "from cartopy import feature as cf\n",
+    "\n",
+    "hv.notebook_extension()\n",
+    "%output widgets='live' size=200"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "<div class=\"alert alert-info\" role=\"alert\">\n",
+    "  <strong>Alert:</strong> The widgets in this notebook a set to render load dynamically and require a live notebook server. If you're viewing this notebook on a static website the widgets won't respond.\n",
+    "</div>"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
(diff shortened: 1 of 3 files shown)


jbednar:
So is there no meaningful way to do it without live and dynamic modes, so that it will be usable (in part) on the website, and then anyone running it themselves can enable live/dynamic locally?

philippjfr:
I could severely downsample the Cube before I do anything with it but that seems to defeat the point of the tutorial, which is to explore the whole dataset. Basically I don't think making it a worse example, just so it works correctly on the website is a good idea. Particularly examples like the Curve are meant to show you can explore the entire dataset in this way, choosing a few limited samples does not convey that.

jlstevens:
I think it looks fine, especially given the alert box at the top.

One suggestion is that you could downsample and use Holomaps for the website, noting in the alert that when working with the live server that the data isn't downsampled. Then you could edit the notebook so it looks like it is using live mode. 

In other words, you could try to fake things so the widgets on the website do *something* but this probably isn't worth the effort, especially as it would require constant manual updates.

doc/Working_with_Iris_Cubes.ipynb
```diff
@@ -0,0 +1,383 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {
+    "collapsed": false
+   },
+   "outputs": [],
+   "source": [
+    "northern = cube.select(latitude=(25, 75))\n",
+    "(northern.select(longitude=(260, 305)).to(gv.Image, geo_dims) *\n",
```
bjlittle: @philippjfr Should be `longitude=(260, 350)` surely ?

philippjfr: The idea was to leave a gap to demonstrate you can overlay subregions of interest even if they are not contiguous.

jbednar:
So there isn't an option that one could set to (a) allow all widgets to work in a static exported HTML copy, and (b) limit the total amount of data exported in such a way?  E.g. can one declare something at the top that makes all widgets later declared as dynamic actually be static, with 3 frames of data (beginning, middle, end of range)?  Seems like that would work as long as the data is randomly accessible (with maybe a different option to just get the first two frames for ones that aren't randomly accessible, like simulations or drawing from a real-time stream).  I find it very frustrating to try to go through an example on the website where I can't even tell what it is supposed to do because none of the widgets work; it just looks like our code is buggy! We can't control where people randomly start looking, and if they start looking at some broken-seeming widgets, they think it's broken.  I wouldn't want to do a bunch of manual hacks to get it working, but it seems like this is a very common need: convey what the notebook does, concretely, in static form, to motivate people to install the live notebook and give them a point of reference for being sure that the live one is working.  It seems like it shouldn't be hard to achieve this, to me, and would make our lives better by making it easier to go back and forth between static and live notebook versions.

PR #21, opened 2016-06-30
Support for geometries, bokeh and xarray

philippjfr:
Apologies I didn't split this into multiple PRs but it's all interdependent. This PR does three main things:

1. Avoids having a hard-dependency on iris, currently needed for image and contour plots but nothing else. If possible at least images should work without iris eventually but that's a later PR.
2. Adds the Polygons and Path types that already exist in HoloViews but adding support to project them to the target coordinate reference system. Additionally adds a ``Shape`` type, which wraps any shapely geometry and renders it in both matplotlib and bokeh. Shapely is a dependency of cartopy and provides very convenient means of loading GeoJSON, shape files and more. It also allows having holes in polygons, which is currently not supported by our Polygons type. 
3. Adds an initial bokeh backend for a subset of the plot types, currently WMTS, Points, Polygons, Path and Shape are supported.


jlstevens:
Looks good and it does make sense to separate the mpl and bokeh backends, under the assumption bokeh will become more capable in future. Hopefully, we can eventually get rid of the iris dependency here and simply make use of the iris cube support in holoviews.

Happy to merge if you think it is ready.

jlstevens:
I understand it is now ready to merge. All this is great stuff but please break it up into more separate PRs next time!

PR #29, opened 2016-08-18
Updated docs to use xarray examples

philippjfr:
Completely overhauls the existing Tutorials to use netCDF data files (exported from iris-sample-data), which can be loaded with either xarray or iris. Still need to define a good way to grab the data from http://assets.holoviews.org/geoviews-sample-data.zip and get a few PRs in HoloViews committed and merged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,15 +27,15 @@ multi-figure layouts of overlaid objects.
 
 ## Installation
 
-You will first probably want to install iris and/or xarray for working with
-multidimensional datasets:
+GeoViews requires cartopy, and most people will want iris and/or
+xarray as well:
 
 ```
 conda install -c scitools/label/dev -c conda-forge iris cartopy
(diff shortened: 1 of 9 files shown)


jbednar:
You could copy the data download script from https://github.com/bokeh/datashader/blob/master/examples/download_sample_data.py ...

philippjfr:
Okay I've now made a number of changes that should make fetching the sample-data easier. If you're working with a dev version you can now simple run:

```python
import geoviews as gv
gv.sample_data()
```

And will fetch the data and put it in the right place for the notebooks to find them. For a released version you'll be able to fetch the notebooks and the sample data using:

```python
import geoviews as gv
gv.examples(include_data=True)
```

which will create a folder containing all the example notebooks and the sample-data. Let me know what you think, using latest HoloViews master this should now all be working.

PR #50, opened 2017-03-14
Update Working_with_Bokeh.ipynb

andreas-h:
add missing parentheses to `coastline()` call in last cell


jbednar:
Thanks!  That's a good point; adding the parentheses does help make sure that the object will have styling independent of previous cells.  I'll go ahead and merge this, though we need to fix the tests (which are failing for an unrelated reason to do with versions of other packages).

andreas-h:
That's interesting.  We were in a situation yesterday where not including the `()` lead to a ValueError complaining about latitude not be in a certain dimension list?!  With the `()` everything went smoothly.

(Sorry I can't be more specific; still very new to Geo/HoloViews, and it was on a colleague's computer)

PR #64, opened 2017-04-19
Add support for MercatorTicker in bokeh

philippjfr:
This PR adds support for the new MercatorTicker in bokeh, mostly addressing this issue: https://github.com/ioam/geoviews/issues/36. Unfortunately it causes various issue when zooming beyond the allowed ranges. I'll have to investigate further to see if bounding the ranges is sufficient to avoid these issues.

### DIFF
--- a/geoviews/plotting/bokeh/__init__.py
+++ b/geoviews/plotting/bokeh/__init__.py
@@ -5,7 +5,7 @@ import param
 import numpy as np
 import shapely.geometry
 from cartopy.crs import GOOGLE_MERCATOR
-from bokeh.models import WMTSTileSource
+from bokeh.models import WMTSTileSource, MercatorTickFormatter, MercatorTicker
 from bokeh.models.tools import BoxZoomTool
 
 from holoviews import Store, Overlay, NdOverlay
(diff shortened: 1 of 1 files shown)


jbednar:
Checking out the lat/lon support has been on my list for a couple of weeks now, but I have yet to try it. :-(

philippjfr:
>Checking out the lat/lon support has been on my list for a couple of weeks now, but I have yet to try it. :-(

You could try this PR, it works well when you're zoomed into some small region but if you try to plot something global, the zoom starts stuttering and the axis ticks mess up pretty badly.

jbednar:
Are there some safeguards that could be put into the Bokeh code to make this safer, so that e.g. ticks just aren't shown when they can't be calculated?

philippjfr:
So I can no longer reproduce many of the issues I had with this initially. The one issue I do still see is that if the plot is larger than the whole globe the ticks do not initialize correctly:

screen shot 2017-09-29 at 7 02 02 pm

However once you zoom in a bit they behave correctly:

screen shot 2017-09-29 at 7 03 24 pm

philippjfr:
I think that behavior is still preferable to having the axis display northings and eastings in meters, which are entirely useless. So I think we should just merge this.

PR #75, opened 2017-06-29
Added geoviews RGB type

philippjfr:
As the title says, adds a geo/crs-aware RGB Element type. Addresses https://github.com/ioam/geoviews/issues/67

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -7,7 +7,8 @@ from cartopy.io.shapereader import Reader
 from holoviews.core import Element2D, Dimension, Dataset as HvDataset, NdOverlay
 from holoviews.core.util import basestring, pd
 from holoviews.element import (Text as HvText, Path as HvPath,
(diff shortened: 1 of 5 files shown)


geoviews/element/geo.py
```diff
@@ -214,6 +215,28 @@ class Image(_Element, HvImage):
     group = param.String(default='Image')
 
 
+class RGB(_Element, HvRGB):
```
jlstevens: Couldn't ``_Element`` have been called ``CartElement`` or similar? It has ``_abstract=True`` so I don't see the need for the underscore...

philippjfr: It could have been yes, but at this point I'll just leave it.

jlstevens: Ok, maybe file an issue for next time.

jlstevens:
Looks fine to me. Merging.

PR #76, opened 2017-06-29
GeoViews types default to PlateCarree CRS

philippjfr:
All geoviews Elements default to PlateCarree CRS.


philippjfr:
Reasoning outlined in: https://github.com/ioam/geoviews/issues/39

jlstevens:
Thanks for linking to the issue. Reasoning makes sense as a change to me so I'll merge.

PR #77, opened 2017-06-29
Small compatibility fix for holoviews 1.8

philippjfr:
As the title says, this was a typo which now clashes with a HoloViews method.


jlstevens:
Merging.

PR #83, opened 2017-08-17
Allow passing bokeh tile source URL to WMTS

philippjfr:
Allows passing tile source URL directly to WMTS constructor avoiding having to import ``WMTSTileSource`` from bokeh. Also maintains backward compatibility so you can continue to pass in instances.

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -24,6 +24,11 @@ try:
 except:
     WMTSTileSource = None
 
+try:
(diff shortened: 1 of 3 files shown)


jlstevens:
Good to see this fixed - this has long been an oddity/annoyance!

jlstevens:
Looks good to me! Merging.

jbednar:
Looks great, thanks!

PR #87, opened 2017-09-14
Add a faster image reprojection implementation

philippjfr:
A faster implementation for image reprojection which will be used by the bokeh interface. Here are some basic profiling results for the old (cartopy regrid) and new (resampling) implementation:

bokeh_plot

### DIFF
--- a/geoviews/operation.py
+++ b/geoviews/operation.py
@@ -7,11 +7,11 @@ from holoviews.core.util import cartesian_product
 from holoviews.operation import Operation
 from shapely.geometry import Polygon, LineString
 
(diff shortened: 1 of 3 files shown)


geoviews/operation.py
```diff
@@ -87,7 +88,68 @@ def _process(self, img, key=None):
                      vdims=img.vdims, crs=proj)
 
 
-class project(ElementOperation):
+
+class project_image_fast(Operation):
```
jlstevens: Seems super useful with great performance! Shame about the name though...

philippjfr:
Going to merge this. Have tested it and tests are passing.

PR #90, opened 2017-09-29
Various optimizations for static sources

philippjfr:
When an Element is static the ``get_data`` call generally still gets called because in some cases the styling options can still affect the returned data. This PR ensures that any data which definitely has not changed isn't computed in the method. This can result in large performance boosts particularly for large Shape, Path and Polygon Elements, which need to be appropriately unpacked.


jbednar:
Looks good to me; ready to merge?

philippjfr:
Depends on https://github.com/ioam/holoviews/pull/1952, so we'll have to make sure to pin the holoviews version on the next release.

philippjfr:
Otherwise ready to merge yes.

jbednar:
Let's merge that HV PR ASAP, then!

PR #101, opened 2017-11-03
Add geoviews.data package to installation

gbrener:
This is to fix an `ImportError` that gets raised by [this line](https://github.com/ioam/geoviews/blob/master/geoviews/__init__.py#L13) in geoviews for the "geoviews.data" package.


philippjfr:
Thanks for getting to this so quickly. Shouldn't affect tests so I'll merge right away.

gbrener:
Sure, no problem. Thanks

PR #102, opened 2017-11-03
Add projection aware bokeh stream callbacks

philippjfr:
This PR will ensure that bokeh stream callbacks appropriately project coordinates back to the original coordinate reference system. As a first step it splits the plotting baseclasses into their own module.


philippjfr:
Tests just passed, merging.

PR #103, opened 2017-11-03
Fixed and disabled fast image projection

philippjfr:
Small fixes for fast project_image option and disabled it by default on bokeh plots.


jbednar:
Why disabled by default?

philippjfr:
Because it's not actually safe, I'll have to follow up in a later PR.

philippjfr:
Okay, tests passing, going to merge and release.

PR #113, opened 2017-12-19
Stopped pinning numpy

philippjfr:


jlstevens:
Maybe there is a *minimum* version worth setting?

philippjfr:
On travis? It'll be determined by the requirements of other libraries, which is why I'm trying to get away from pinning something.

philippjfr:
Okay, installing everything from conda-forge seems to addressed the issues here, merging.

PR #116, opened 2018-01-05
Added projection aware QuadMesh element

philippjfr:
This PR adds a projection aware QuadMesh element which allows plotting gridded data with irregular sampling and/or multi-dimensional coordinates. This is fairly slow when using the bokeh backend so it would be good to address proper datashader GeoViews integration soon, which should get around these issues.

Addresses: https://github.com/ioam/geoviews/issues/57, https://github.com/ioam/geoviews/issues/73 and https://github.com/ioam/geoviews/issues/81


philippjfr:
Now passing, will merge.

PR #120, opened 2018-01-13
Unified WMTS element support across backends

philippjfr:
The WMTS element now supports the tile source URL format that is also supported by bokeh, finally unifying the two. This PR also deprecates ``Tiles`` since it does exactly what WMTS does. I'd also be happy to keep Tiles as a simple alias for WMTS though, @jbednar any thoughts?

Here's the same plot in two backends:

```
tiles = gv.WMTS('https://maps.wikimedia.org/osm-intl/{Z}/{X}/{Y}@2x.png') 

# Project data to Web Mercator
nybb = gpd.read_file(gpd.datasets.get_path('nybb'))
poly_data = nybb.to_crs(ccrs.GOOGLE_MERCATOR.proj4_init)
polys = gv.Polygons(poly_data, vdims='BoroName', crs=ccrs.GOOGLE_MERCATOR)
tiles * polys
```

image

screen shot 2018-01-13 at 10 07 00 pm

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -2,7 +2,7 @@ import param
 import numpy as np
 from cartopy import crs as ccrs
 from cartopy.feature import Feature as cFeature
-from cartopy.io.img_tiles import GoogleTiles as cGoogleTiles
+from cartopy.io.img_tiles import GoogleTiles
 from cartopy.io.shapereader import Reader
 from holoviews.core import Element2D, Dimension, Dataset as HvDataset, NdOverlay
 from holoviews.core.util import (basestring, pd, max_extents,
@@ -32,7 +32,7 @@ except:
 
 from ..util import path_to_geom, polygon_to_geom
 
-geographic_types = (cGoogleTiles, cFeature, BaseGeometry)
+geographic_types = (GoogleTiles, cFeature, BaseGeometry)
 
 def is_geographic(element, kdims=None):
     """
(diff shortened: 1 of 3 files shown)


jbednar:
Great!  The name WMTS has always been mysterious-looking, and I do think a better name would be good.  Is Tiles the appropriate name?  Are there WMTS sources that are *not* map tiles?  If so it's fine; if not MapTiles might be a better name.  I also like TileSource or TileServer, because the one thing that's important to know about Tiles/WTMS objects is that they are served by an external site, unlike just about any other Element.

philippjfr:
``TileSource`` seems reasonable, that said it would add a third alias, with the other two names (Tiles and WMTS) having to go through a fairly long deprecation cycle, so maybe I favor ``Tiles``.

>Are there WMTS sources that are not map tiles?

Don't think so since WMTS stands for Web Mapping **Tile** Service.

philippjfr:
I'm going to go ahead and merge once tests pass, please file an issue to add a ``TileSource`` and deprecate WMTS and Tiles if you think it's a better name.

jbednar:
> Misread your question, I think the XYZ scheme uses the assumption that it's map tiles in mercator coordinates, so MapTiles might be more accurate.

Right; my question was confusing, but what I was getting at was whether there are tile sources that our objects would accept that are not maps.  Bokeh seems to allow for the possibility, in that it has a [TileSource base class](https://bokeh.pydata.org/en/latest/docs/reference/models/tiles.html#bokeh.models.tiles.TileSource) that does not mention maps.  I believe that tiles have been used in Bokeh for spectrogram displays that are not maps but I suppose can somehow be made to fit into the XYZ format (as long as Y is unbounded, as it is time), and I want to make sure that (a) we don't rule out applications like that, and (b) we have names that correctly reflect the level of generality.

jbednar:
(This being geoviews, we'd have to add a Tiles object to hv to handle such cases, presumably!)

PR #122, opened 2018-01-14
Added geographic VectorField element

philippjfr:
Straightforward addition of a geographic ``VectorField`` element which simply reuses project_points to do the projecting and the holoviews VectorFieldPlot implementations.


jlstevens:
Neat!

PR #130, opened 2018-02-06
Fixed element type hierarchies

philippjfr:
All GeoViews elements should appropriately subclass their HoloViews counterparts.


philippjfr:
Push build passed.

PR #132, opened 2018-02-06
Switch to holoviews RasterPlot

philippjfr:
The ImagePlot class in holoviews' mpl backend was orphaned and deleted so GeoViews should use the RasterPlot class.


philippjfr:
PR build passing.

PR #142, opened 2018-03-09
created stand-alone scripts that reproduce the plots in the docs

scaine1:
When I first started not having working examples was quite difficult. It's obvious to me now how to get the same results as the ipython magics, but at the time I recall it was quite confusing 

These should be useful for people who don't or can't use the notebooks.

I tried to show a few different ways to set the plot options, i.e. globally, or locally, using dictionary or strings. 

I wasn't 100% sure when to put them, so ended up putting them in the notebooks/user_guide directory.

In the docs there is a link that says "Right click to download this notebook from GitHub."

I would also suggest adding "Right click to download this script from GitHub."

### DIFF
--- a/notebooks/user_guide/Geometries.ipynb
+++ b/notebooks/user_guide/Geometries.ipynb
@@ -75,8 +75,17 @@
    "outputs": [],
    "source": [
     "%%output backend='bokeh'\n",
-    "%%opts Feature [scale='110m']\n",
-    "gf.ocean * gf.land() * gv.Feature(graticules, group='Lines').opts(plot=dict(scale='110m'))"
+    "%%opts Feature.Land.110m [scale='110m']\n",
+    "%%opts Feature.Land.50m [scale='50m']\n",
+    "(gf.ocean * gf.land().relabel(label='110m') * gv.Feature(graticules, group='Lines') + \n",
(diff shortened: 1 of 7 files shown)


jbednar:
Thanks! These are clearly useful for GeoViews users, but I'd be concerned about our ability to maintain them and keep them up to date as the original versions change over time.  For one thing, as we implement https://github.com/ioam/geoviews/pull/121 , we'll have dozens of separate notebooks, each of which could usefully have an associated script.

Should we not just use an automatic mechanism to convert each notebook into a runnable python script, which I believe @jlstevens has prototyped if I am not making that up?  Then every page could have a link saying "Right click to download this code as a **notebook** or as a **script** to try it out yourself".

jlstevens:
> Should we not just use an automatic mechanism to convert each notebook into a runnable python script, which I believe @jlstevens has prototyped if I am not making that up? 

That's right although it is (as you say) a prototype right now though if I have a chance I can clean it up (the conversion will be done using the holoviews console script).

scaine1:
The inclusion of the .options method recently already made me think maintaining will be an issue. Having a conversion script would be heaps better!
I was making the scripts as an exercise anyway so no issue if you want to wait till conversion tool comes out.

PR #143, opened 2018-03-12
Switch versioning system to use autover

jlstevens:
PR to switch geoviews to autover, following the corresponding PR in https://github.com/ioam/holoviews/pull/2426. Embedding version.py until it is available via param.

The only thing I'm not entirely sure about is whether a MANIFEST will need to be added for the ``.version`` file.


jlstevens:
I have released 1.5.0a1 [here](https://anaconda.org/ioam/geoviews/files) using this PR (and pushed the tag). As far as I can tell, the tests are only failing because of a few flake issues.

jlstevens:
@philippjfr I think this is ready to review/merge when the tests pass and when you confirm that the alpha works as expected.

jlstevens:
I think this PR is now ready to review/merge.

.gitattributes
```diff
@@ -0,0 +1,2 @@
+ __init__.py export-subst
+setup.py export-subst
```
philippjfr: Newlines here and in .gitignore would be nice, but won't let it hold up the PR.

philippjfr:
One minor comment about newlines, I'll let you decide what to do, happy to merge either way.

.gitattributes
jlstevens: Addressed in 3efb09f

jlstevens:
Addressed your comment and the tests are green.

PR #154, opened 2018-04-24
Updated geoviews to use param.Version

jlstevens:
Follows how holoviews does it and allows ``version.py`` to be removed.


setup.py
```diff
@@ -44,11 +48,14 @@ def get_setup_version(reponame):
     import json, importlib
```
philippjfr: This ``importlib`` import no longer appears to be used and is causing flake errors.

jlstevens: Thanks. Must be a difference in flake8 versions as this wasn't caught in holoviews? I'll fix it in both places now.

ceball: (it's a difference in how flake8 is run)

philippjfr: Right, in holoviews it's only run on the holoviews and tests packages.

ceball: When I inflict pyct on you, it will be the same everywhere, and everyone will hate me ;)

jlstevens:
Looks like the tests will pass this time.

PR #168, opened 2018-05-13
Added is_global option to set global plot extent

philippjfr:


geoviews/plotting/bokeh/plot.py
```diff
@@ -96,6 +99,8 @@ def get_extents(self, element, ranges):
         set_extent method to project the extents to the
         Elements coordinate reference system.
         """
+        if self.is_global:
+            return (-20026376.39, -20048966.10, 20026376.39, 20048966.10)
```
jbednar: I don't quite get the reason for the is_global flag (some description in the PR would be useful!), but in any case, how does knowing it's a plot of the whole globe determine the range here?  Can't there be infinitely many plots of the whole globe, centered at different longitudes?

philippjfr: It's modeled on the cartopy ``ax.set_global()`` method, which simply means "Set the extent of the Axes to the limits of the projection."

>Can't there be infinitely many plots of the whole globe, centered at different longitudes?

No, neither matplotlib nor currently bokeh support wrapping at arbitrary longitudes. I have no issue adding a ``central_longitude`` option at a later point but that will require a whole bunch more work (and I'm not yet sure that it's even possible for certain projections in matplotlib).

philippjfr: Actually I take the part about matplotlib back, it is possible to set the ``central_longitude`` in matplotlib directly on the ``crs`` in some cases.

jlstevens:
Just as a matter of taste, I think ``is_global`` is a pretty horrible name that doesn't really convey what it does. Why not just ``global`` or ``globe``?

philippjfr:
I don't particularly disagree, ``globe`` is imo wrong, and global cannot be used because ``global = ...`` produces a syntax error.

jlstevens:
Why do you think ``globe`` is wrong? Maybe I don't know exactly what this parameter is doing but it *sounds* like it should you the whole globe (assuming the extents of projections go that far?). To be honest, something like ``show_globe`` or ``show_global`` might be more consistent with holoviews (though I'm not such a fan of these ``show_X`` methods).

philippjfr:
globe in the context of cartopy refers specifically to projections that actually render the earth as a globe, which means that part of it is obscured. Not a fan of ``show_...`` either but ``show_global`` is probably the best option here.

philippjfr:
Maybe ``global_extent``?

PR #181, opened 2018-05-23
Added back some direct dependencies.

ceball:
When I cleaned up the dependencies, I removed some dependencies that are direct dependencies of geoviews. I'm not sure why I did that. I think the direct dependencies should all be listed. What do you think?

This PR shouldn't make any practical difference anywhere (numpy and param being dependencies of hv).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,6 @@ stages:
     if: tag =~ ^v(\d+|\.)*[^a-z]\d*$    
   - name: docs
     if: tag =~ ^v(\d+|\.).*$
-  # TODO: add release packages (future PR)
 
 jobs:
   include:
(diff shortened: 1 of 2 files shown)


jbednar:
I think it's clearer to list numpy and param explicitly as you do here, even if there's no chance HoloViews would ever not have them as dependencies, and thus no chance they won't already be covered.

philippjfr:
I have no opinion but we might as well add them.

ceball:
@jbednar, you might once have pointed me to a tool for listing what's imported by a set of notebooks and/or python modules, but I can't remember what it was. I think maybe it was a tool by @gbrener? I think it could be useful for pyct, so if someone could point me to it again, I'd be grateful.

gbrener:
https://github.com/gbrener/python-utils/blob/master/import_utils.py

jbednar:
I'd be happy to have that functionality made available in pyct somewhere...

PR #183, opened 2018-05-24
Tiling updates

jbednar:
* Replaced "ESRI" with "EsriImagery" in examples
* Removed "ESRI" from the list of default tile sources
* Fixed missing attribution for Stamen Watercolor and incorrect attribution for Stamen Terrain (based on current advice from [Stamen](http://maps.stamen.com/#howto))
* Removed confusing statement in docs that made it seem like the Matplotlib backend did not support web tile servers


philippjfr:
Looks fine to me, should I merge?

PR #188, opened 2018-06-07
Removed references to non-existent notebooks.

ceball:
The `rm ...` in travis is why the doc build failed for the last tag push.

This reminds me...I can't remember if the doc build for geoviews is doing what you want now:
  * push dev tag: updates https://ioam-docs.github.io/geoviews-dev/
  * push release tag: updates https://ioam-docs.github.io/geoviews-release/

You could build the docs locally, too, if you were working on the docs specifically `(*)`. (Could also set it up so pushing a special tag, e.g. 'doc-build' or whatever, updates another site e.g. https://ioam-docs.github.io/geoviews-tmp/ or something like that.)

`(*)` Not quite there yet, but there will be a simple nbsite command to run locally for that soon, I hope. But for the purposes of deciding if the options cover what you want, please pretend it exists.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,7 +79,6 @@ jobs:
         - pip install git+https://github.com/pyviz/nbsite.git
         - geoviews fetch-data --path=examples
         - bokeh sampledata
-        - rm examples/Colocated_cube_pair.ipynb examples/CubeExplorer_Prototype.ipynb # not currently part of website
         # note: will vastly simplified in a future version of nbsite
         - cd doc
         - nbsite_nbpagebuild.py ioam geoviews ../examples .
--- a/tox.ini
+++ b/tox.ini
@@ -67,8 +67,6 @@ addopts = -v --pyargs --doctest-modules --doctest-ignore-import-errors
 norecursedirs = doc .git dist build _build .ipynb_checkpoints
 # depend on optional iris, xesmf, etc
 nbsmoke_skip_run = .*Homepage\.ipynb$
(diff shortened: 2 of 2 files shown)


ceball:
Actually, while on this topic, a separate but related question is about versioning the docs. 

The docs are versioned in that they are "standalone" (a single, self-contained directory that just needs to be served) and have the version number "inside them":

screen shot 2018-06-07 at 9 00 30 am

However, (a) that's not very obvious, and (b) we don't keep around previous versions. 

I've always imagined doing something like what bokeh does, unless that doesn't seem good to you? On your hosting provider for geoviews, could you do something along the lines of uploading the docs to a folder like `1.4/`, `1.5/`, '1.6/`, etc, and then update `latest` point to the latest one of those (and have geoviews.org coming from latest)?

If that seems reasonable, we could make nbsite support that kind of workflow.

jlstevens:
> On your hosting provider for geoviews, could you do something along the lines of uploading the docs to a folder like 1.4/, 1.5/, 1.6/, etc, and then update latest point to the latest one of those (and have geoviews.org coming from latest)?

> If that seems reasonable, we could make nbsite support that kind of workflow.

I would very much like this for all our projects. The main disadvantage I see is the increased storage costs on S3 as the number of versions grow. Unless we have a policy of having the docs only ever going back two versions (say)?

philippjfr:
Yes we should definitely start versioning docs. I can start doing the uploads to our S3 buckets that way, but it would also be nice to add a toggle to switch between the different versions like bokeh has.

PR #192, opened 2018-06-14
Initial version of roadmap

jbednar:
Anyone have modifications or additions?

Presumably this and the one on PyViz should be linked into the docs somewhere, but I'm not sure where.  Any suggestions?

### DIFF
--- a/doc/Roadmap.rst
+++ b/doc/Roadmap.rst
@@ -1,0 +1,73 @@
+GeoViews Roadmap, as of 6/2018
+==============================
+
+GeoViews is under continuous development and improvement, with
+priorities determined by user input, the needs of current projects,
+and external funding.
+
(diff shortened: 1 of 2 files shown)


jbednar:
Any opinion on where to link this in?  Possibilities:

1. In the main TOC, after User Guide and Gallery and before About
2. Buried as a link under About.
3. In the same list as Gitter, Twitter, and Github.

I vote for 1, but that does make it quite prominent, so others may prefer a less prominent spot.

jbednar:
You can see an example of approach 1 at https://ioam-docs.github.io/pyviz-latest .

philippjfr:
I'm fine with 1.

jbednar:
Ok, I've added a link to the roadmap, so I think this is ready to merge.

PR #198, opened 2018-07-06
Packaging fixups

ceball:
* Prevent installation of old (pre-split) geoviews alongside geoviews-core.
* Switch from host to build (equivalent when there's no building, but host is recommended).

### DIFF
--- a/conda.recipe/core/meta.yaml
+++ b/conda.recipe/core/meta.yaml
@@ -18,6 +18,8 @@ build:
     {% endfor %}
 
 requirements:
+  run_constrained:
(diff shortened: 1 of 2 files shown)


philippjfr:
Thanks, let me know when it's ready to merge. I'll tag 1.5.2 later today.

ceball:
It's ready to merge.

By the way, before releasing, I would do a 1.5.2a0 tag first to make sure that everything works. If everything (tests, packages, docs) turns out ok, then I would tag 1.5.2 (same commit as the successful alpha tag).

PR #199, opened 2018-07-06
Simplify travis file (should be no change to behavior).

ceball:
The idea is that gradually more stuff is turning into env vars, then to config of pyctdev...


ceball:
Would be good to have a second pair of eyes check the changes (obscure/dull though they are), but apart from that, should be ready to merge.

(appveyor tests haven't run yet, but aren't relevant)

philippjfr:
Looks good thanks! Can't say I'm likely to spot any issues at a glance though.

PR #216, opened 2018-08-13
Ensure that WMTS url is updated correctly

philippjfr:
Partial fix for https://github.com/pyviz/geoviews/issues/215

Despite the fact that the URL is now updated correctly, the tile source does not seem to update indicating an issue in bokeh.


jlstevens:
I can confirm that this branch reflects changes to the URL if you keep zooming in to a map (if you don't zoom in, you just see the cached version from the initial URL).

PR #221, opened 2018-08-16
Add RGB.load_tiff classmethod

philippjfr:
Adds a RGB.load_tiff classmethod which attempts to infer the crs from the data.


jbednar:
Neat!

PR #225, opened 2018-08-30
Handle negative resolution when loading tiff

philippjfr:
Ensures that negative resolutions flip the coordinates inside a tiff, ensuring that the image displays with the correct orientation by default.

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -333,11 +333,11 @@ class RGB(_Element, HvRGB):
         with warnings.catch_warnings():
             warnings.filterwarnings('ignore')
             da = xr.open_rasterio(filename)
-        return cls.from_tiff(da, crs, apply_transform, **kwargs)
+        return cls.from_xarray(da, crs, apply_transform, **kwargs)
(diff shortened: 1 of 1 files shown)


jlstevens:
Thanks!

One more general issue (that may not be in the scope of this PR) is that when using tools like `intake`, you don't get a 'tif' file to load. This is a bit of a pain as the things that `load_tiff` does (transfer crs, apply `res`, fix coordinate ordering) are really useful and it would be nice to have similar functionality when starting with an `xarray` data structure.

philippjfr:
I've now added factored out a ``.from_xarray`` methods, which will unpack an xarray DataArray appropriately.

jlstevens:
Looks great! My only question is whether you think there is a difference between ``from_X`` and ``load_X``. Why not ``load_xarray`` to be consistent?

philippjfr:
Because it doesn't load any xarray file, only ones that can be loaded with ``xr.open_rasterio`` and I'm not quite sure what else xr.open_rasterio works on.

PR #231, opened 2018-09-21
Switch to pyviz main channel for pyctdev on CI

ceball:
I.e. only used released versions of pyctdev.


ceball:
Should be safe to merge - just means geoviews will keep using the same version that it's using currently, rather than some potentially disruptive dev releases before the next main release.

PR #245, opened 2018-10-22
Add projected grid lines to gv feature

ahuang11:
I was thinking of adding it to bokeh/plot.py/GeoPlot, but I didn't know how to do it without making the resulting element a GeoOverlayPlot automatically so I thought maybe just make it more accessible as a feature

image


geoviews/plotting/bokeh/plot.py
```diff
@@ -65,6 +65,12 @@ def __init__(self, element, **params):
             show_bounds = self._traverse_options(element, 'plot', ['show_bounds'],
                                                  defaults=False)
             self.show_bounds = not any(not sb for sb in show_bounds['show_bounds'])
+            if self.show_grid:
```
philippjfr: We could consider having show_grid add these gridlines in future.

philippjfr:
Thanks, this looks like a valuable addition. I've made one comment but we can address that in future.

PR #282, opened 2019-01-18
Add geometry resampling operation

philippjfr:
This operation makes it possible to dynamically work with relatively large geometry datasets. The ``resample_geometry`` operation is similar to other resampling operations such as ``rasterize``, ``regrid`` and ``decimate`` in that it progressively adjusts the detail of the input element as you zoom in and out of a plot. I'll let the docstring explain how it works:

```
    This operation dynamically culls and resamples Path or Polygons
    elements based on the current zoom level. On first execution a
    RTree created using the Sort-Tile-Recursive algorithm, which is
    used to query for geometries within the current viewport (defined
    by the x_range and y_range).

    Any geometries returned by the RTree query are tested to ensure
    their area is over the display_threshold, expressed as a fraction
    of the current viewport area. Any remaining polygons are
    simplified using the Douglas-Peucker algorithm, which eliminates
    vertices while ensuring that the curve does not diverge from the
    original curve by more than the tolerance. The tolerance is
    expressed as a fraction of the square root of the area of the
    current viewport. 

    Once computed a simplified geometry is cached depending on the
    current zoom level. The number of valid zoom levels can be
    declared and are used to recursively subdivide the domain into
    smaller subregions.

    If requested the geometries can also be clipped to the current
    viewport which avoids having to render vertices that are not
    visible.
```

(See the [wikipedia entry](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm) for more details).

As a simple example here is the operation applied to the land geometry feature:

```python
resample_geometry(gf.land.geoms('50m')).opts(width=600, height=300, projection=ccrs.PlateCarree())
````

land_world

```python
resample_geometry(gf.coastline.geoms('10m')).opts(width=600, height=300, projection=ccrs.PlateCarree())
```

coastlines

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -139,6 +139,43 @@ class Feature(_GeoFeature):
     def __call__(self, *args, **kwargs):
         return self.opts(*args, **kwargs)
 
+    def geoms(self, scale=None, bounds=None, as_element=True):
+        """
+        Returns the geometries held by the Feature.
+
+        Parameters
+        ----------
+        scale: str
+           Scale of the geometry to return expressed as string, e.g.
+           '10m', '50m' or '110m'
+        bounds: tuple
+           Tuple of a bounding region to query for geometries in
+        as_element: boolean
+           Whether to wrap the geometries in an element
+
+        Returns
+        -------
+        geometries: Polygons/Path
+           Polgons or Path object wrapping around returned geometries
+        """
+        feature = self.data
+        if scale is not None:
+            feature = feature.with_scale(scale)
+
+        if bounds:
+            extent = (bounds[0], bounds[2], bounds[1], bounds[3])
+        else:
+            extent = None
+        geoms = [g for g in feature.intersecting_geometries(extent) if g is not None]
+        if not as_element:
+            return geoms
+        elif not geoms or 'Polygon' in geoms[0].geom_type:
+            return Polygons(geoms, crs=feature.crs)
+        elif 'Point' in geoms[0].geom_type:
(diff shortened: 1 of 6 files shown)


geoviews/element/geo.py
```diff
@@ -139,6 +139,43 @@ def __init__(self, data, kdims=None, vdims=None, **params):
...
 
+    def geoms(self, scale=None, bounds=None, as_element=True):
+        """
+        Returns the geometries held by the Feature.
+
+        Parameters
+        ----------
+        scale: str
+           Scale of the geometry to return expressed as string, e.g.
+           '10m', '50m' or '110m'
```
jbednar: How does a user know what scales are available?

geoviews/operation/resample.py
```diff
@@ -0,0 +1,197 @@
...
+    display_threshold = param.Number(default=0.0001, doc="""
+        The fraction of the current viewport covered by a geometry
+        before it is shown.""")
+
+    dynamic = param.Boolean(default=True, doc="""
+       Enables dynamic processing by default.""")
+
+    preserve_topology = param.Boolean(default=False, doc="""
+        Whether to preserve topology between geometries
+        (much faster when disabled).""")
```
jbednar: Can you explain what the tradeoff is?  What does it mean to preserve topology?

philippjfr: I'll update the docstring, basically it avoids self-intersecting or otherwise invalid geometries but is **much** slower.

Issue #291, opened 2019-01-29
Use new nbsite syntax


jsignell:
Clean up build commands in travis  - adding to dodo.py as needed and using new nbsite syntax.

PR #292, opened 2019-01-29
Using new nbsite syntax and new doit command

jsignell:
* Switched to new nbsite syntax for generate_rst and build
* Added data_setup command to dodo
* Using new data_setup command on travis, appveyor
* Stopped using `ioam` channel

Spot tested locally that gallery works:
screen shot 2019-01-29 at 10 44 10 am

And json blob is copied (tried dragging slider):
screen shot 2019-01-29 at 10 45 09 am

Closes: #291

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -85,8 +85,8 @@ jobs:
         - cd doc
         - ln -s ../examples/assets assets
         - ln -s ../examples/data data
-         - nbsite_gallery.py .. examples None '.' bokeh matplotlib None
-         - rm gallery/**/*.ipynb
+        - nbsite_gallery.py .. examples None '.' bokeh matplotlib None
+        - rm gallery/**/*.ipynb
         - cd ..
         ### build
         - nbsite build --what=html --output=builtdocs
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,8 +17,8 @@ env:
   global:
     - PYENV_VERSION=3.6
     - PKG_TEST_PYTHON="--test-python=py36 --test-python=py27"
-    - CHANS_DEV="-c pyviz/label/dev -c ioam"
-    - CHANS_REL="-c pyviz -c ioam"
+    - CHANS_DEV="-c pyviz/label/dev"
+    - CHANS_REL="-c pyviz"
     - LABELS_DEV="--label dev"
     - LABELS_REL="--label dev --label main"
     - MPLBACKEND="Agg"
@@ -38,7 +38,7 @@ stages:
   - name: conda_package
     if: tag =~ ^v(\d+|\.)*[^a-z]\d*$
   - name: pip_package
-    if: tag =~ ^v(\d+|\.)*[^a-z]\d*$    
+    if: tag =~ ^v(\d+|\.)*[^a-z]\d*$
 
 jobs:
   include:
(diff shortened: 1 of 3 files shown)


philippjfr:
Just tagged a dev release, will see how that works out.

philippjfr:
Looks like it failed when deploying the docs:

```
Deploying application
Initialized empty Git repository in /tmp/d20190129-8039-186hf2p/work/.git/
Switched to a new branch 'gh-pages'
cd -
cd /tmp/d20190129-8039-186hf2p/work
rsync: change_dir "/home/travis/build/pyviz/geoviews/doc/_build/html" failed: No such file or directory (2)
rsync error: some files/attrs were not transferred (see previous errors) (code 23) at main.c(1183) [sender=3.1.0]
Could not copy /home/travis/build/pyviz/geoviews/doc/_build/html.
failed to deploy
```

I'll update the ``local_dir`` in the deploy section.

jsignell:
I think I misunderstood what I was checking for when running the gallery script. I thought the screenshots were being dynamically created.

philippjfr:
>I think I misunderstood what I was checking for when running the gallery script. I thought the screenshots were being dynamically created.

They are, the thumbnails are created correctly. The issue is that the notebooks themselves aren't executed so the output is just the raw code.

jsignell:
The evaluated notebooks need to be in doc/gallery/

philippjfr:
>The evaluated notebooks need to be in doc/gallery/

I don't follow, sphinx should evaluate the notebooks, the problem here is that the gallery script probably copies the notebooks to doc/gallery and nbsite then assumes that they have already been evaluated.

jsignell:
If you want to run each notebook twice, then you could just delete the .ipynb files in doc/gallery after the thumbnailing has completed.

philippjfr:
>If you want to run each notebook twice, then you could just delete the .ipynb files in doc/gallery after the thumbnailing has completed.

The thumbnailing does not execute the notebook it converts to script and runs that, I think the original intention of copying was that the cleared notebooks would be included in the build so they can be downloaded.

jsignell:
Right then all you need is to delete the .ipynb files from /doc/gallery after the thumnailing finishes. The next command will build them.

philippjfr:
>Right then all you need is to delete the .ipynb files from /doc/gallery after the thumnailing finishes.

Sure that's an easy fix now, since I'm guessing they are not included anyway, but the idea here is that they are eventually shipped as part of the docbuild so a user can download the notebook from the website.

.travis.yml
```diff
@@ -74,29 +73,27 @@ jobs:
...
-        - cd doc
-        - nbsite_nbpagebuild.py ioam geoviews ../examples .
+        - doit data_setup
+        ### generate rst skeleton
+        - nbsite generate-rst --org pyviz --project-name geoviews --offset 0
         ### gallery
+        - cd doc
         - ln -s ../examples/assets assets
         - ln -s ../examples/data data
         - nbsite_gallery.py .. examples None '.' bokeh matplotlib None
```
jsignell:
```suggestion
         - nbsite_gallery.py .. examples None '.' bokeh matplotlib None
         - rm gallery/**/*.ipynb
```

philippjfr:
Thanks, should work now.

philippjfr:
Superseded by #298

PR #296, opened 2019-02-10
Add VectorField/WindBarbs project operation

### LINKED ISSUE #295: VectorField does not reproject angle
Same data, two different projections (PlateCarree and NorthPolarStereo). Note how the arrows are the same, just their locations changes. The obvious fix is to reproject before calculating magnitude and angle but it might be nice to mention this in the documentation.

Cartopy on the other hand (further below) seems to do this (using `quiver`), so it might be easy to implement (then, however, using East/North velocities instead of magnitude/angle:

platecarree

npstereographic

cartopy1

cartopy2

philippjfr:
- [x] Fixes #295

### DIFF
--- a/geoviews/operation/projection.py
+++ b/geoviews/operation/projection.py
@@ -157,7 +157,7 @@ class project_shape(_project_operation):
 
 class project_points(_project_operation):
 
-    supported_types = [Points, Nodes, VectorField, HexTiles, Labels]
+    supported_types = [Points, Nodes, HexTiles, Labels]
 
     def _process_element(self, element):
         if not len(element):
@@ -194,8 +194,6 @@ class project_geom(_project_operation):
     supported_types = [Rectangles, Segments]
 
     def _process_element(self, element):
-        if not len(element):
-            return element.clone(crs=self.p.projection)
         x0d, y0d, x1d, y1d = element.kdims
         x0, y0, x1, y1 = (element.dimension_values(i) for i in range(4))
         p1 = self.p.projection.transform_points(element.crs, x0, y0)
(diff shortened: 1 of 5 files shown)


poplarShift:
As usual, terrific response time!

I believe you made a mistake in the calculation of us, vs from the angle, see my commit: https://github.com/pyviz/geoviews/compare/project_vectors...poplarShift:project_vectors

Your fix seems fine otherwise. It does change the values of the angles though that get fed into the color mapping, so it's not ideal, but a good step in the right direction.

Geoviews:
gv_wgs

gv_np

Cartopy:
cartopy_wgs

cartopy_np

philippjfr:
Thanks for the fixes. I think the coloring issue is to do with the magnitudes being transformed but the color range being computed before the transformation. I'll think about how best to fix that.

ahuang11:
```python
import numpy as np
import geoviews as gv
import matplotlib.pyplot as plt
import cartopy.crs as ccrs
import panel as pn
from metpy.calc import wind_components, wind_direction
from metpy.units import units

gv.extension("matplotlib")
pn.extension()

x = np.linspace(10, 50, 3)
X, Y = np.meshgrid(x, x)
U, V = 10 * X, 5 * Y
angle = np.pi / 2 - np.arctan2(-V, -U)
magnitude = np.hypot(U, V)
```

```python
fig = plt.figure()
ax = plt.axes(projection=ccrs.Orthographic())

ax.quiver(X, Y, U, V, transform=ccrs.PlateCarree())
ax.set_global()
ax.coastlines()

vectorfield = (
    gv.VectorField(
        (X, Y, angle, magnitude), label="Vector Field", crs=ccrs.PlateCarree()
    ).opts(
        global_extent=True,
        projection=ccrs.Orthographic(),
        magnitude="Magnitude",
        show_legend=False,
    )
    * gv.feature.coastline()
)
pn.Row(
    pn.pane.Matplotlib(fig, dpi=180, width=400, height=400),
    pn.panel(vectorfield, width=300, height=300),
)
```

```python
fig = plt.figure(figsize=(6, 6))
ax = plt.axes(projection=ccrs.Orthographic())
ax.barbs(
    X,
    Y,
    U,
    V,
    transform=ccrs.PlateCarree(),
    linewidth=0.5,
    sizes=dict(emptybarb=0.25, spacing=0.2, height=0.5),
)
ax.set_global()
ax.coastlines()

windbarbs = (
    gv.WindBarbs((X, Y, angle, magnitude), crs=ccrs.PlateCarree()).opts(
        global_extent=True,
        projection=ccrs.Orthographic(),
        show_legend=False,
        linewidth=0.5,
        sizes=dict(emptybarb=0.25, spacing=0.2, height=0.5),
        fig_size=200,
    )
    * gv.feature.coastline()
)

windbarbs_uv = gv.WindBarbs.from_uv((X, Y, U, V), crs=ccrs.PlateCarree()).opts(
    global_extent=True,
    projection=ccrs.Orthographic(),
    show_legend=False,
    linewidth=0.5,
    sizes=dict(emptybarb=0.25, spacing=0.2, height=0.5),
    fig_size=200,
    )

pn.Row(
    pn.pane.Matplotlib(fig, dpi=300, width=400, height=400),
    pn.panel(windbarbs, width=400, height=400),
    pn.panel(windbarbs_uv, width=400, height=400),
)
```

image
image

ahuang11:
I think this is ready for review.

Hoxbro:
When I look at the figures, there seem to be differences in Matplotlibs and this implementation as we move closer to the border of the globe, see:
image

Do you know why?

geoviews/operation/projection.py
```diff
@@ -222,6 +220,55 @@ def _process_element(self, element):
         return element.clone(tuple(new_data[d.name] for d in element.dimensions()),
                              crs=self.p.projection)
 
+class project_vectorfield(_project_operation):
+
+    supported_types = [VectorField]
+
+    def _calc_angles(self, ut, vt):
+        # mathmetical convention
```
Hoxbro, line 228:
```suggestion
        # mathematical convention
```
Maybe also worth noting that this is convention is chosen because of matplotlib if I recall correctly. 

Same with project_windbarbs.

ahuang11:
> When I look at the figures, there seem to be differences in Matplotlibs and this implementation as we move closer to the border of the globe, see: image
> 
> Do you know why?

I suspect it's due to:
```

pivot{'tail', 'mid', 'middle', 'tip'}, default: 'tail'
The part of the arrow that is anchored to the X, Y grid. The arrow rotates about this point.

'mid' is a synonym for 'middle'.
```

ahuang11:
Setting pivot='tail'
```python

vectorfield = (
    gv.VectorField(
        (X, Y, angle, magnitude), label="Vector Field", crs=ccrs.PlateCarree()
    ).opts(
        global_extent=True,
        projection=ccrs.Orthographic(),
        magnitude="Magnitude",
        show_legend=False,
        pivot="tail",
    )
    * gv.feature.coastline()
)
pn.Row(
    pn.pane.Matplotlib(fig, dpi=180, width=400, height=400),
    pn.panel(vectorfield, width=300, height=300),
)
```
Results in the same plot:
image

ahuang11:
All set.

PR #321, opened 2019-05-05
Support hover lon/lat for other projs

ahuang11:
Fixes:
https://github.com/pyviz/geoviews/issues/251

Linked with:
https://github.com/pyviz/holoviews/pull/3692

image

I tried adding to kdim, but the param is bounded between (2, 2)...


philippjfr:
Yeah, I'm -1 on this implementation, particularly because it doesn't actually handle data that's already in one of the unsupported projection. We could consider un-projecting at plot level to add the columns but that has potentially significant performance implications.

PR #332, opened 2019-06-03
Matplotlib axis has no property gridlines (bugfix)

poplarShift:
Self-explanatory. Unsure if this used to be part of the matplotlib API, but `.gridlines()` is definitely not the way to set axis grids now.


ahuang11:
https://github.com/pyviz/geoviews/pull/308/files Quite similar :)

poplarShift:
Haha, yeah I was wondering why one of the two _finalize_axis routines had the right one. I think I'd looked into this some months ago but never got around to making the PR.

philippjfr:
Merged #351 which appears to have fixed the issue.

PR #351, opened 2019-08-16
Add grid labels

ahuang11:
https://github.com/pyviz/geoviews/issues/350

I went for the try: except TypeError method because it seems cleaner and probably faster than checking three types (platecarree, mercator, and google mercator)

Not sure if this is the best way to make xyticks show

### DIFF
--- a/geoviews/plotting/mpl/__init__.py
+++ b/geoviews/plotting/mpl/__init__.py
@@ -2,8 +2,10 @@ import copy
 
 import numpy as np
 import param
+import matplotlib.ticker as mticker
 from cartopy import crs as ccrs
 from cartopy.io.img_tiles import GoogleTiles
+from cartopy.mpl.gridliner import LONGITUDE_FORMATTER, LATITUDE_FORMATTER
 
 try:
     from owslib.wmts import WebMapTileService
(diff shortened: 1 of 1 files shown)


geoviews/plotting/mpl/__init__.py
```diff
@@ -104,11 +106,47 @@ def __init__(self, element, **params):
...
+                gl.ylabels_left = False
+
+            gl.yformatter = LATITUDE_FORMATTER
+
     def _finalize_axis(self, *args, **kwargs):
         ret = super(GeoPlot, self)._finalize_axis(*args, **kwargs)
         axis = self.handles['axis']
-        if self.show_grid:
-            axis.gridlines()
+        try:
```
ahuang11: There's no if because xticks/yticks require the gridlines

geoviews/plotting/mpl/__init__.py
```diff
@@ -104,11 +106,47 @@ def __init__(self, element, **params):
...
+
+            gl.xformatter = LONGITUDE_FORMATTER
+
+        if self.yaxis and self.yaxis != 'bare':
+            if isinstance(self.yticks, list):
+                gl.ylocator = mticker.FixedLocator(self.yticks)
+            elif isinstance(self.yticks, int):
+                gl.ylocator = mticker.MaxNLocator(self.yticks)
+
+            if self.yaxis in ['left', 'right-bare']:
```
ahuang11: Couldn't find a way to draw actual tick marks so they become equal...

philippjfr:
Appears to be working, thanks!

PR #377, opened 2019-11-04
Port custom tools, streams and links from EarthSim

philippjfr:


jbednar:
Presumably these will need some user guide notebooks?  I also don't see what makes them geoviews specific (as opposed to HoloViews) on a quick glance.

philippjfr:
>I also don't see what makes them geoviews specific (as opposed to HoloViews) on a quick glance.

I realize this discussion happened a while ago but we discussed the plan that these would move to GeoViews for the time being on multiple occasions and eventually **could** migrate up to HoloViews under two conditions: 

a) We decide to generalize the annotators further 
b) We decide we actually want to ship bokeh extensions as part of HoloViews itself

jbednar:
@philippjfr and I have been discussing this, and I believe the plan is (which Philipp can correct if needed):

- For now, move the tools here into GeoViews so that they become available for use immediately
- All tools except for PolyVertexEditTool are useful for any Bokeh drawing-tools user, and we will try to get those into Bokeh itself.
- PolyVertexEditTool would stay here in GeoViews indefinitely, as it's a specialized geo-specific tool
- HoloViews won't need any new tools and thus will continue not to need any Bokeh extensions.
- The main documentation and examples will be in HoloViews, with some geo-specific examples here.

philippjfr:
@kcpevey You might also want to test this PR which adds geo-support for the annotators along with some of the other custom tools.

philippjfr:
I'll merge for now so I can test releasing the custom bokeh models.

PR #395, opened 2019-11-30
Add codecov

ceball:
Mainly I'm triggering a commit to see if I've fixed codecov.io for hvplot.

Anyone know if we need this file? I copied it from panel.

(Same question for al lthe repos that are using codecov)


ceball:
Hmm, confusing: the badge is for codecov.io but the repo is set up for coveralls.

PR #410, opened 2019-12-19
Fix stats on coveralls.io

ceball:
coveralls.io appears to be confused about coverage. We speculate it might be from multiple steps uploading coverage, even though they do not measure coverage.

This PR avoids uploading no coverage from various build steps.

We could do this in a more elegant way, but let's at least see if this works first.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,6 +61,7 @@ jobs:
     - <<: *default
       env: DESC="py2 flakes" PYTHON_VERSION=2.7
       script: doit test_flakes
+      after_success: true # don't upload coverage
 
     ########## DOCS ##########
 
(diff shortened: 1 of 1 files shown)


philippjfr:
Thanks for trying, I'll merge when tests pass.

ceball:
(The travis push build has been failing for me even just with trivial doc changes - not sure what that's about.)

philippjfr:
Presumably you hadn't been rebased?

philippjfr:
That does potentially look like a bug though.

ceball:
I've rebased now. (But I think this also happened to me earlier on another PR where I was up to date with master. Having said that, you work so fast I could easily be wrong 😂 )

ceball:
I rebased and the tests passed (you were right), but I don't get what the new conflict is about - doesn't look like something that would be a problem for git to merge itself...

ceball:
Oh I see, github just hadn't updated the situation on my screen when I wrote my comment (there was a force push on master). I'll rebase again.

philippjfr:
>Oh I see, github just hadn't updated the situation on my screen when I wrote my comment (there was a force push on master)

I'll stop doing that I promise, environment issues have a habit of making me adopt desperate measures.

ceball:
Ha ha, no problem - I know that feeling :)

PR #424, opened 2020-01-09
Implement Geometry conversion for remaining geometry elements

philippjfr:
Adds geometry conversions using the geom method for Points, Rectangles and Segments and then consistently demonstrates conversions in Annotators user guide.


jbednar:
Cool! Do you think that showing the shapely object is really necessary that many times in the manual? I worry that it makes it a bit long, but if you think people are likely to miss it if it's only on one annotator, sure.

philippjfr:
I think it's good to see that it is consistently supported.

Issue #446, opened 2020-03-24
Conda Install Fails


jjoonathan:
`conda install -c pyviz geoviews`  currently fails to Solve Environment inside a clean anaconda container launched with `sudo docker run -it continuumio/anaconda3`. 

I've tried the pyviz channel, the default channel, and the conda-forge channel with the same results.

Conda suggests that the problem has to do with conflicts and that it is looking for incompatibilities, but it has been running since yesterday and only reports 90% progress. I'll update this post if it completes.

`Found conflicts! Looking for incompatible packages.
This can take several minutes.  Press CTRL-C to abort.
Examining conflict for ...`

jjoonathan:
Here is conda info inside the container:
```bash
sudo doc(base) jon@jon-Precision-7720:~/hyperspectral$ sudo docker run -it continaconda3 
(base) root@68c9563607c9:/# conda info

     active environment : base
    active env location : /opt/conda
            shell level : 1
       user config file : /root/.condarc
 populated config files : 
          conda version : 4.8.2
    conda-build version : 3.18.11
         python version : 3.7.6.final.0
       virtual packages : __glibc=2.28
       base environment : /opt/conda  (writable)
           channel URLs : https://repo.anaconda.com/pkgs/main/linux-64
                          https://repo.anaconda.com/pkgs/main/noarch
                          https://repo.anaconda.com/pkgs/r/linux-64
                          https://repo.anaconda.com/pkgs/r/noarch
          package cache : /opt/conda/pkgs
                          /root/.conda/pkgs
       envs directories : /opt/conda/envs
                          /root/.conda/envs
               platform : linux-64
             user-agent : conda/4.8.2 requests/2.22.0 CPython/3.7.6 Linux/5.3.0-42-generic debian/10 glibc/2.28
                UID:GID : 0:0
             netrc file : None
           offline mode : False
```

jjoonathan:
`conda install -c pyviz geoviews-core` completes successfully. It's just the full geoviews that conda has issues with. I wish it didn't take so long (~an hour) to fail, so that I could iterate quicker, but that's how conda is.

Update: conda's dependency analysis finally finished after 18 hours! Unfortunately it does not look terribly helpful at first glance. Hopefully it's more helpful to one of you :)
[conda_conflict_report.txt](https://github.com/holoviz/geoviews/files/4376394/conda_conflict_report.txt)

jjoonathan:
[...]

PR #447, opened 2020-03-25
Added conda update --all to conda install instructions

jjoonathan:
I've just added `conda update --all` with minimal explanation here, as keeping the install instructions streamlined is a top priority and running an update before an install should be a familiar concept with a familiar risk profile in most cases. This particular form of the update command is recommended here: https://www.anaconda.com/keeping-anaconda-date/ and I've verified that it solves the issue here: https://github.com/holoviz/geoviews/issues/446 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,14 +46,22 @@ objects.
 
 ## Installation
 
-You can install GeoViews and its dependencies using conda (to ensure you have a recent version of conda run `conda update -n root conda` first):
+If you want the latest GeoViews, you will need an up-to-date environment. Updating is never risk-free, but it is a good idea in general and the commands `conda list --revisions` and `conda install --revision N` can usually recover from updates gone awry.
 
+```
+conda update --all
+```
+
+You can then install GeoViews and all of its dependencies with the following:
+
+```
+conda install -c pyviz geoviews
+```
 
 Alternatively you can install the geoviews-core package, which
 only installs the minimal dependencies required to run geoviews:
 
 ```
-conda update --all
 conda install -c pyviz geoviews-core
 ```
 
--- a/README.md
+++ b/README.md
@@ -46,12 +46,8 @@ objects.
 
 ## Installation
 
-You can install GeoViews and its dependencies using conda:
+You can install GeoViews and its dependencies using conda (to ensure you have a recent version of conda run `conda update -n root conda` first):
 
-```
-conda update --all
-conda install -c pyviz geoviews
-```
 
 Alternatively you can install the geoviews-core package, which
 only installs the minimal dependencies required to run geoviews:
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ objects.
 You can install GeoViews and its dependencies using conda:
 
 ```
+conda update --all
 conda install -c pyviz geoviews
 ```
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -49,13 +49,15 @@ objects.
 You can install GeoViews and its dependencies using conda:
 
 ```
+conda update --all
 conda install -c pyviz geoviews
 ```
```
philippjfr:
```suggestion
You can install GeoViews and its dependencies using conda (to ensure you have a recent version of conda run `conda update -n root conda` first):

```
conda install -c pyviz geoviews
```
```

philippjfr:
Thanks for your PR. I'm no 100% sure if we should always recommend users run `conda update --all` though, in most cases I would expect it to be sufficient to update conda itself.

jjoonathan:
Yeah, there are definitely tradeoffs. On one hand, it's all but guaranteed to mess up someone's environment at some point, but on the other hand, not doing it is all but guaranteed to mess up someone's environment at some point, too (both the "rotting anaconda install" problem, which is common, and the "latest tagged anaconda is buggy" problem, as seen in my bug thread, which I hope is rare).

Three things make me lean towards including it. First, I've seen it before. "Install you updates! Eat your broccoli!" are relatively common occurrences at the top of instructions all over the web. Second, the semantics are pretty clear. It says what it does, and I'd expect someone who went in with the intention of avoiding updates to figure out that they should omit it. Finally, anaconda seems to have blessed `conda update --all` as [the standard update mechanism](https://www.anaconda.com/keeping-anaconda-date/). Even though one might assume that sticking to pinned releases would be standard, conda not only doesn't recommend it, they actually make it quite difficult to do. You might think that `conda update anaconda` updates to the next numbered release, but it doesn't -- as far as I can tell, the only way to stick with numbered releases is to search out and specify anaconda release numbers, which doesn't seem to be something that people commonly recommend doing, let alone actually do.

I almost forgot the fourth reason: the current "numbered" release is flat out broken on linux and the maintainers don't seem to be in a hurry to fix it, while `conda update --all` fixes it right now :-)

In any case, for reference, here are the candidate "update" commands and their semantics:

* `conda update conda`  updates the conda package manager but not any other package. It would not fix the dependency issue above.
* `conda update anaconda` one might expect it to update to the latest "tagged" anaconda metapackage, but in fact it updates to the latest (consistent) release of every package in the anaconda metapackage, which is almost guaranteed to move the user off of a "tagged" release if they are on one (see above).
* `conda update --all` updates every package to the latest (consistent) release, regardless of whether or not it is in the Anaconda metapackage.  Blessed as the standard update mechanism: https://www.anaconda.com/keeping-anaconda-date/

The final decision, of course, is yours, and I wouldn't blame you for deciding against the PR, it's very much a two edged sword.

jbednar:
Hmm. `conda update --all` seems very dangerous to me, and if that's what's needed for a user to get things working, a safer suggestion would seem to be to tell them to create a brand-new environment just for using geoviews; that way we couldn't be blamed for messing up their current environment.  If we had that instruction, it's not one I would use personally; it would cause needless churn and uncertainty.

Is there a smaller change that would fix the specific problem this is addressing?  E.g. removing the `anaconda` metapackage?

jjoonathan:
Well, both removing the metapackage and updating it with `conda update anaconda` are likely to create side effects. How about breaking `conda update --all` into a separate section that explicitly advises about the risks?

philippjfr:
Sorry this sat for 9 months!!!

PR #472, opened 2020-09-21
Handle proj4.js API change in Hover formatter

philippjfr:


rsignell-usgs:
@philippjfr, if this passes, is there a chance to get a minor release pushed out? 
I would much rather conda install geoviews!  😸

philippjfr:
Yep working on a 1.8.2 now.

PR #473, opened 2020-09-21
Bump geoviews.js version and bokeh requirements

philippjfr:


rsignell-usgs:
@philippjfr , looks like maybe we just need to conda install nodejs to avoid these failures?

Issue #512, opened 2021-07-28
gv.feature.states fill_alpha should default to 0


andhuang-CLGX:
Else, it blocks the view

jbednar:
It's a breaking change in behavior, but I agree it makes sense, and it looks like you were the one to add it (https://github.com/holoviz/geoviews/pull/312), so I'd support such a PR.

ahuang11:
Haha forgot it was me who added it!

PR #513, opened 2021-08-03
Make states fill_alpha = 0

ahuang11:
https://github.com/holoviz/geoviews/issues/512


geoviews/plotting/bokeh/__init__.py
```diff
@@ -310,5 +310,6 @@ def _process(self, element, key=None):
 options.Feature.Ocean  = Options('style', fill_color='#97b6e1', line_color='#97b6e1')
 options.Feature.Lakes  = Options('style', fill_color='#97b6e1', line_color='#97b6e1')
 options.Feature.Rivers = Options('style', line_color='#97b6e1')
+options.Feature.States = Options('style', fill_alpha=0)
```
jbednar, line 313: I guess Borders is a polyline rather than a polygon, and thus unfilled?  Looks like Land is filled by default, but I guess that makes more sense. So I guess this is the one change that's needed!

Issue #509, opened 2021-07-20
Expose the `"extra_url_vars"` param from bokeh


andhuang-CLGX:
specially, I have a time variable too for radar imagery

https://docs.bokeh.org/en/latest/_modules/bokeh/models/tiles.html

andhuang-CLGX:
I want to parameterize the TIME variable with a pn.widgets.DiscreteSlider

```
from bokeh.models.tiles import BBoxTileSource
from bokeh.plotting import figure
from bokeh.tile_providers import get_provider, Vendors
from bokeh.io import show

p = figure(x_range=(-2000000, 6000000), y_range=(-1000000, 7000000),
           x_axis_type="mercator", y_axis_type="mercator")
url = "https://nowcoast.noaa.gov/arcgis/rest/services/nowcoast/analysis_meteohydro_sfc_rtma_time/MapServer/export?transparent=true&format=png8&layers=show%3A7&bboxSR=3857&imageSR=3857&size=256,256&f=image&time={TIME}&bbox={XMIN},{YMIN},{XMAX},{YMAX}"
tile = BBoxTileSource(url=url, extra_url_vars={"TIME": "1627412400000"})
p.add_tile(tile)
show(p)
```

PR #515, opened 2021-08-03
Expose bokeh tiles kwds

ahuang11:
Not entirely sure if more work is needed to expose the variables from:
https://docs.bokeh.org/en/latest/docs/reference/models/tiles.html
https://github.com/holoviz/geoviews/issues/509

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -223,6 +223,21 @@ class WMTS(_GeoFeature):
 
     layer = param.String(doc="The layer on the tile service")
 
+    extra_url_vars = param.Dict(doc="""
+        A dictionary that maps url variable template keys to values.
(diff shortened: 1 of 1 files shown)


philippjfr:
Yeah, this PR doesn't actually do anything in it's current state. You'll actually have to pass those parameters down to the underlying bokeh renderer.

philippjfr:
Oh also, these parameters should **not** live on the element, they should be exposed on the plotting class.

philippjfr:
Indeed, at least `min_zoom` and `max_zoom` are already valid options you can set using `.opts`.

philippjfr:
Will close since the PR basically has to be redone from scratch. Sorry it took me a while to look at this.

PR #539, opened 2021-11-16
Add missing _process_msg method to GeoPolyEditCallback

maximlt:
`GeoPolyEditCallback` was the only callback whose `_process_msg` was not defined. This is going to be useful for the GrabCut example (https://github.com/pyviz-topics/examples/pull/193).


philippjfr:
Will merge the OSX 3.8 Fiona error can be fixed later.

Issue #465, opened 2020-06-18
support lower case WMTS url template variables


scottyhq:
#### ALL software version info
geoviews=1.8.1
holoviews=1.13.2
bokeh=2.0.1

#### Description of expected behavior and the observed behavior
gv.WMTS() docs specify "The WMTS Element represents a Web Map Tile Service specified as a
    URL containing {x}, {y}, and {z} templating variable". Many servers provide urls with lowercase template variables, but it seems geoviews throws and error unless upper case characters are used. I expect both lower case and upper case to work.

#### Complete, minimal, self-contained example code that reproduces the issue

```
import geoviews as gv
gv.extension('bokeh')
gv.WMTS('https://gibs.earthdata.nasa.gov/wmts/epsg3857/best/BlueMarble_NextGeneration/default/EPSG3857_500m/{z}/{y}/{x}.jpeg')
```

#### Stack traceback and/or browser JavaScript console output
```pytb
/srv/conda/envs/notebook/lib/python3.7/site-packages/geoviews/plotting/bokeh/__init__.py in get_data(self, element, ranges, style)
     56             tile_source = WMTSTileSource
     57         else:
---> 58             raise ValueError('Tile source URL format not recognized. '
     59                              'Must contain {X}/{Y}/{Z}, {XMIN}/{XMAX}/{YMIN}/{YMAX} '
     60                              'or {Q} template strings.')

ValueError: Tile source URL format not recognized. Must contain {X}/{Y}/{Z}, {XMIN}/{XMAX}/{YMIN}/{YMAX} or {Q} template strings.
```

maximlt:
Fixed by https://github.com/holoviz/geoviews/pull/542

PR #542, opened 2021-11-18
Accept lowercase xmin/ymin/xmax/ymax

ahuang11:
https://github.com/holoviz/geoviews/issues/465

image


jbednar (approved):
Changes look good, thanks. The test failures look like they have a common source, e.g. some change in xarray, but I don't know what's happening there.

maximlt:
They're mostly, if not entirely, due to changes in shapely 1.8.0. Will try to fix that in the next days.

PR #543, opened 2021-11-23
Adapt to shapely 1.8.0 and 2.0

maximlt:
Shapely 1.8.0 has recently been released and is unfortunately not compatible with geoviews. The most critical issue is that geoviews relied on the `array_interface_base`, which unfortunately contained a bug in shapely 1.8.0 (see this currently open PR https://github.com/Toblerity/Shapely/pull/1214) and forces users of geoviews to downgrade shapely.

shapely 1.8.0 prepares the next major version bump of shapely towards 2.0. They've release a *Migrating* guide which has been very useful (https://shapely.readthedocs.io/en/stable/migration.html).

### DIFF
--- a/geoviews/data/geom_dict.py
+++ b/geoviews/data/geom_dict.py
@@ -136,7 +136,7 @@ class GeomDictInterface(DictInterface):
         if isinstance(geom, Polygon) and geom.interiors:
             return True
         elif isinstance(geom, MultiPolygon):
-            for g in geom:
+            for g in geom.geoms:
                 if isinstance(g, Polygon) and g.interiors:
                     return True
         return False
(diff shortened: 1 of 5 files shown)


jbednar (commented):
Looks good, assuming that the change from `geom` to `geom.geoms` is backwards compatible. I've no idea what the significance of the shift from `np.bool` to `np.bool_` is.

maximlt:
The change from `geom` to `geom.geoms` is some API clean-up done by shapely. `.geoms` has been there forever (well, not quite so, it might not predate Param!). I'll have a closer look at the difference between `np.bool` and `np.bool_` though.

philippjfr:
Looks good, please finish up the `np.bool_` changes and then feel free to merge yourself.

maximlt:
About `np.bool` vs. `np.bool_` vs. `bool`, Numpy has deprecated the usage of `np.bool` in its [1.20.0 release](i https://numpy.org/doc/stable/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated). They explain that `np.bool` was an alias for the builtin `bool` which was confusing users. In the release notes they indicate that:

> To give a clear guideline for the vast majority of cases, for the types `bool`, `object`, `str` (and `unicode`) using the plain version is shorter and clear, and generally a good replacement.

I think that despite their recommendation I will go for `np.bool_`, since the type of each element really is a `np.bool_` object and not a pure Python boolean object.

image

maximlt:
Unrelated test failure on MacOs (fiona failng due to an error in cython apparently https://github.com/Toblerity/Fiona/issues/1043)

PR #546, opened 2021-12-01
CI: pin fiona and matplotlib on py38 and macos

maximlt:
Pushed to `master` a first patch by mistake.


maximlt:
It seems like I had to pin `matplotlib` **and** `fiona` to get the test suite to pass on macos. Merging then and will prepare a PR to unpin that in the future when the issue is solved elsewhere.

PR #548, opened 2021-12-22
Fix typo

maximlt:
Introduced in #514


jbednar (approved):
Thanks for catching that!

maximlt:
The test failure is unrelated to the changes brought by this PR.

Let's keep on eye on that though, we've had to add a hack in the CI of holoviews to get it to work on MacOS (without being able to find the root cause of the issue in a reasonable amount of time), if the issue here happens here to be the same we'll have to find what's going wrong with conda/pyctdev/doit/...

Issue #522, opened 2021-09-10
gv.RGB does not work with NxMx3 arrays


TheoMathurin:
#### ALL software version info
Minimal example run with the following config, but encountered this issue in other environments:
- geoviews 1.9.1
- holoviews 1.14.5
- bokeh 2.3.3
- datashader 0.13.0
- cartopy 0.18.0
- python 3.9.7
- jupyter and bokeh server
- CentOS 7 / RHEL 7 / Ubuntu 18.04

#### Description of expected behavior and the observed behavior
In contrast to `hv.RGB`, `gv.RGB` only takes input arrays with shape NxMx4. Sending a NxMx3 array will raise an error at rendering, while [the docstrings](https://github.com/holoviz/geoviews/blob/master/geoviews/element/geo.py) states that _"[An] input array may have a shape of NxMx4 or NxMx3. In the latter case, the defined alpha dimension parameter is appended to the list of value dimensions."_

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import numpy as np
import geoviews as gv
import cartopy.crs as ccrs

gv.extension('bokeh')

img = np.array(np.zeros((100, 100, 3)))
coastline = gv.feature.coastline
rgb = gv.RGB(img, bounds=(-10, -10, 10, 10), crs=ccrs.PlateCarree()))

# When trying to render the following (in a notebook or bokeh server), an error will be issued
(rgb*coastline).opts(projection=ccrs.PlateCarree(), width=600, height=300)
```

#### Stack traceback and/or browser JavaScript console output

Last element of traceback:
```
~/miniconda3/envs/jupyter/lib/python3.9/site-packages/holoviews/core/data/image.py in values(cls, dataset, dim, expanded, flat, compute, keep_index)
    197             # than expanded column format, reorient before expanding
    198             if dataset.data.ndim > 2:
--> 199                 data = dataset.data[:, :, dim_idx-dataset.ndims]
    200             else:
    201                 data = dataset.data

IndexError: index 3 is out of bounds for axis 2 with size 3
```

PR #558, opened 2022-02-17
Fix handling of 3-channel RGB element

philippjfr:
The `hv.RGB` constructor adds the alpha channel if needed so it should not be declared in the parameter definition.

Fixes https://github.com/holoviz/geoviews/issues/522


philippjfr:
Damn, thanks for reporting that. We'll have to follow up with another release.

PR #570, opened 2022-05-02
Adapt to shapely 2.0

maximlt:
Warnings seen while running hvplot's notebooks. Continue the work started in https://github.com/holoviz/geoviews/pull/543. This warning was not seen as this code path doesn't seem to be run part of the CI.

### DIFF
--- a/geoviews/operation/projection.py
+++ b/geoviews/operation/projection.py
@@ -76,7 +76,7 @@ class project_path(_project_operation):
             if isinstance(geom, Polygon) and geom.area < 1e-15:
                 continue
             elif isinstance(geom, MultiPolygon):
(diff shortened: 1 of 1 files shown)


TheoMathurin:
Hi,

Does this solve this other Shapely-related deprecation warning, popping up when calling `gv.operation.resample_geometry`?

`ShapelyDeprecationWarning: STRtree will be changed in 2.0.0 and will not be compatible with versions < 2.
  tree = STRtree(geoms)`

maximlt:
Hmm I ran the test suite to list the warnings triggered by Shapely. `resample_geometry` isn't tested apparently so I missed this deprecation.

PR #578, opened 2022-07-06
Fix building with pip install

Hoxbro:
I noticed that `python -m pip install .` did not work and raised the following traceback because `setuptools` was not present in `pyproject.toml`. 

<details>
  <summary>Traceback </summary>

``` python-traceback
ERROR: Exception:
Traceback (most recent call last):
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/cli/base_command.py", line 167, in exc_logging_wrapper
    status = run_func(*args)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/cli/req_command.py", line 205, in wrapper
    return func(self, options, args)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/commands/install.py", line 341, in run
    requirement_set = resolver.resolve(
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/resolver.py", line 75, in resolve
    collected = self.factory.collect_root_requirements(root_reqs)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 506, in collect_root_requirements
    req = self._make_requirement_from_install_req(
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 468, in _make_requirement_from_install_req
    cand = self._make_candidate_from_link(
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/factory.py", line 215, in _make_candidate_from_link
    self._link_candidate_cache[link] = LinkCandidate(
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 291, in __init__
    super().__init__(
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 161, in __init__
    self.dist = self._prepare()
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 230, in _prepare
    dist = self._prepare_distribution()
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/resolution/resolvelib/candidates.py", line 302, in _prepare_distribution
    return preparer.prepare_linked_requirement(self._ireq, parallel_builds=True)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/operations/prepare.py", line 428, in prepare_linked_requirement
    return self._prepare_linked_requirement(req, parallel_builds)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/operations/prepare.py", line 497, in _prepare_linked_requirement
    dist = _get_prepared_distribution(
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/operations/prepare.py", line 58, in _get_prepared_distribution
    abstract_dist.prepare_distribution_metadata(
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/distributions/sdist.py", line 48, in prepare_distribution_metadata
    self._install_build_reqs(finder)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/distributions/sdist.py", line 118, in _install_build_reqs
    build_reqs = self._get_build_requires_wheel()
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/distributions/sdist.py", line 95, in _get_build_requires_wheel
    return backend.get_requires_for_build_wheel()
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_internal/utils/misc.py", line 685, in get_requires_for_build_wheel
    return super().get_requires_for_build_wheel(config_settings=cs)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_vendor/pep517/wrappers.py", line 172, in get_requires_for_build_wheel
    return self._call_hook('get_requires_for_build_wheel', {
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_vendor/pep517/wrappers.py", line 332, in _call_hook
    raise BackendUnavailable(data.get('traceback', ''))
pip._vendor.pep517.wrappers.BackendUnavailable: Traceback (most recent call last):
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 89, in _build_backend
    obj = import_module(mod_path)
  File "/home/shh/miniconda3/envs/holoviz/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 972, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 984, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'setuptools'
```

</details>

After adding it, I could successfully install `geoviews`, though I noticed that a name field was added to `package-lock.json`, so I added it to this commit. Other packages from holoviz seem to have a name like `@holoviz/geoviews` and not only `geoviews`. Should this also be done here?

At last I added `.bokeh` and `node_modules` to `.gitignore`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -6,3 +6,4 @@ requires = [
     "pyviz_comms >=0.6.0",
     "setuptools",
 ]
+build-backend = "setuptools.build_meta"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,5 +4,5 @@ requires = [
     "pyct >=0.4.4",
     "bokeh >=2.4.0,<2.5",
     "pyviz_comms >=0.6.0",
-    "setuptools"
+    "setuptools",
 ]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,5 +3,6 @@ requires = [
     "param >=1.9.0",
     "pyct >=0.4.4",
     "bokeh >=2.4.0,<2.5",
-    "pyviz_comms >=0.6.0"
+    "pyviz_comms >=0.6.0",
+    "setuptools"
 ]
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,5 @@ geoviews/.version
 .doit.db
 .pytest_cache
 
+node_modules/
+.bokeh
--- a/geoviews/package-lock.json
+++ b/geoviews/package-lock.json
@@ -1,11 +1,11 @@
 {
-  "name": "geoviews",
+  "name": "@holoviz/geoviews",
   "version": "1.9.5",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
-      "name": "geoviews",
+      "name": "@holoviz/geoviews",
       "version": "1.9.5",
       "license": "BSD-3-Clause",
       "dependencies": {
--- a/geoviews/package-lock.json
+++ b/geoviews/package-lock.json
@@ -5,6 +5,7 @@
   "requires": true,
   "packages": {
     "": {
+      "name": "geoviews",
       "version": "1.9.5",
       "license": "BSD-3-Clause",
       "dependencies": {
--- a/geoviews/package.json
+++ b/geoviews/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "geoviews",
+  "name": "@holoviz/geoviews",
   "version": "1.9.5",
   "description": "Simple, concise geographical visualization in Python",
   "license": "BSD-3-Clause",
--- a/setup.py
+++ b/setup.py
@@ -145,15 +145,6 @@ extras_require={
 
 extras_require['all'] = sorted(set(sum(extras_require.values(), [])))
 
-# until pyproject.toml/equivalent is widely supported; meanwhile
-# setup_requires doesn't work well with pip. Note: deliberately omitted from all.
-extras_require['build'] = [
-    'param >=1.9.2',
-    'pyct >=0.4.4',
-    'bokeh >=2.4,<2.5',
-    'pyviz_comms >=0.6.0'
-]
-
 
 ########################
 ### package metadata ###


maximlt:
> Other packages from holoviz seem to have a name like @holoviz/geoviews and not only geoviews. Should this also be done here?

That's a good question, @jlstevens do you know what's best? I would tend to think that the `name` key wasn't there for a reason but it would be good to know why.

What `pip` version do you use? I thought `setuptools` was picked up as the default build backend but I'm certainly wrong (Python packaging is a little confusing isn't it).

Also, does `pip install geoviews` lead to the same error? I assume it doesn't but just to double check.

pyproject.toml
```diff
@@ -3,5 +3,6 @@ requires = [
     "param >=1.9.0",
     "pyct >=0.4.4",
     "bokeh >=2.4.0,<2.5",
-    "pyviz_comms >=0.6.0"
+    "pyviz_comms >=0.6.0",
+    "setuptools"
```
maximlt, line 7:
```suggestion
    "setuptools",
```

maximlt, line 7: `setuptools` would also have to be included in the `build` extra group declared in `setup.py`. As I understand it, the  build dependencies declared in `pyproject.toml` are picked up by pip while the build dependencies declared in `setup.py` are to be picked up by pyctdev when you run `doit develop_install`.

Hoxbro:
Yes, python packaging is confusing... I had `setuptools` installed in the environment, but it only succeeded when I added it to `pyproject.toml`. 

My versions are `pip=22.1.2` and `setuptools=62.6.0`.

I can install geoviews with pip, though I first need to install `pyproj` and `cartopy` from `conda`.

maximlt:
> I had setuptools installed in the environment, but it only succeeded when I added it to pyproject.toml.

Yep I think that these are dependencies that are installed in a virtual environment by pip on the build step, so it doesn't matter if they're installed in working environment.

> I can install geoviews with pip install geoviews

That's good news, it would have been really scary if not! Still worth fixing, in particular for those who would like to re-package GeoViews building it from its source.

What do you think about adding `build-backend = "setuptools.build_meta"` to the pyproject.toml file? This seems what to be recommended by [setuptools](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html). Without `build-backend` defined [pip falls back to some other config](https://pip.pypa.io/en/stable/reference/build-system/pyproject-toml/#fallback-behaviour).

pyproject.toml
maximlt: Had a quick look at `pyctdev`'s code and actually it seems that it gets the build dependencies from the pyproject.toml file. So the `build` group extra may just be legacy. It may be worth confirming that.

Hoxbro:
I don't have a strong opinion about adding `build-backend = "setuptools.build_meta"`, though I think it should be consistent through all holoviz packages.

maximlt:
What we could do as an experiment is in this PR:
* add `build-backend` to the pyproject.toml file
* remove the `'build'` extra group in the setup.py file

Then see if the tests still pass, and manually run the build and docs workflow in dry-run mode from Github's UI, to see if they also still pass.

philippjfr:
Agree with @maximlt's suggestions. Also agree that the name in package.json should be `@holoviz/geoviews`. The fact that it's not in package-lock.json is probably just down to an older version of node.js but it should be added there. Once we change the name we may also have to update the NPM token unless it just references the same token that's used by Panel.

Hoxbro:
I have updated the name. Another one needs to check the NPM token.

maximlt:
Ah it's unfortunate the workflows can't be launched manually from GitHub's site when the branch is on a fork :/

PR #579, opened 2022-07-14
Fix building with pip install

Hoxbro:
Same as #578, but moved the branch to the main repo and away from my fork.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,5 @@ geoviews/.version
 .doit.db
 .pytest_cache
 
+node_modules/
+.bokeh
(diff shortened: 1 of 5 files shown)


maximlt:
I've had to re-add the `'build'` extra group in the `setup.py` file, it's needed by the conda recipe:
https://github.com/holoviz/geoviews/blob/937869415526f69098be119c3078e0331c761c4e/conda.recipe/core/meta.yaml#L23-L25

I imagine there's a way for conda-build to pull the dependencies from the pyproject.toml file, I'd keep that for some later work on updating pyctdev.

PR #581, opened 2022-07-15
Fix index outside bounds

Hoxbro:
Fixes https://github.com/holoviz/hvplot/issues/738

It will still raise an error but is much more descriptive. 

Added `from None` to avoid `During handling of the above exception, another exception occurred` message.

### DIFF
--- a/geoviews/util.py
+++ b/geoviews/util.py
@@ -106,7 +106,7 @@ def project_extents(extents, src_proj, dest_proj, tol=1e-6):
                              'to %s projection. Ensure the coordinate '
                              'reference system (crs) matches your data '
                              'and the kdims.' %
-                             (src_name, dest_name))
+                             (src_name, dest_name)) from None
     else:
         geom_in_crs = boundary_poly.intersection(domain_in_src_proj)
     return geom_in_crs.bounds
--- a/geoviews/util.py
+++ b/geoviews/util.py
@@ -99,7 +99,7 @@ def project_extents(extents, src_proj, dest_proj, tol=1e-6):
             geom_in_src_proj = geom_clipped_to_dest_proj
         try:
             geom_in_crs = dest_proj.project_geometry(geom_in_src_proj, src_proj)
-        except ValueError:
+        except (ValueError, IndexError):
             src_name =type(src_proj).__name__
             dest_name =type(dest_proj).__name__
             raise ValueError('Could not project data from %s projection '
--- a/geoviews/tests/test_util.py
+++ b/geoviews/tests/test_util.py
@@ -1,0 +1,19 @@
+import cartopy.crs as ccrs
+import pytest
+
+from geoviews.util import project_extents
+
+
+
+def test_outside_extents():
+    # extents outside bounds of map 32635, see https://epsg.io/32635
+    extents = (-2036817.4174174175, 577054.6546546547, -1801982.5825825825, 867745.3453453453)
+    dest_proj = ccrs.Mercator()
+    src_proj = ccrs.epsg(32635)
+
+    msg = (
+        "Could not project data from .+? projection to .+? projection\. "
+        "Ensure the coordinate reference system \(crs\) matches your data and the kdims\."
+    )
+    with pytest.raises(ValueError, match=msg):
+        project_extents(extents, src_proj, dest_proj)


geoviews/util.py
```diff
@@ -99,14 +99,15 @@ def project_extents(extents, src_proj, dest_proj, tol=1e-6):
...
-                             (src_name, dest_name))
+        except (ValueError, IndexError):
+            src_name = type(src_proj).__name__
+            dest_name = type(dest_proj).__name__
+            raise ValueError(
+                f'Could not project data from {src_name} projection '
+                f'to {dest_name} projection. Ensure the coordinate '
+                'reference system (CRS) matches your data and the kdims, '
+                'and the data is not ouside the bounds of the CRS.'
+            ) from None
```
jbednar, line 110: Given that this is a viz tool designed to support data exploration, isn't there an argument for simply warning about data that can't be projected into these coordinates, while projecting the remainder of the data? E.g. if people have data extending to the poles and use a projection not defined for the full extent, seems like they should be able to see the rest of the data anyway. Otherwise they'll have to figure out the extent of this projection and figure out how to clip out that data, right?

Hoxbro:
This error can occur with the current implementation for coordinates close to the bounds but still inside the bounds.

<details>

<summary> Code used for the examples </summary>

``` python
import xarray as xr
import cartopy.crs as ccrs
import panel as pn
import geoviews as gv
import numpy as np

pn.extension()
gv.extension("bokeh", logo=False)

tiles = gv.tile_sources.Wikipedia().opts(width=1000, height=500, padding=0.1)
crs = ccrs.epsg(32635)

def plot(y1):
    da = xr.DataArray(np.arange(10000).reshape(100,100),
                      coords=dict(x=np.linspace(166021, 166021 + 3000, 100),
                                  y=np.linspace(y1, 3000, 100))
                     )
    return gv.Image(da, crs=crs) * tiles


w = pn.widgets.IntSlider(value=0, start=-2000, end=2000, step=100)
pn.Column(w, gv.DynamicMap(pn.bind(plot, w)))



da = xr.DataArray(np.arange(10000).reshape(100,100),
                  coords=dict(x=np.linspace(166021, 166021 + 3000, 100),
                              y=np.linspace(-3000, 0, 100)))
crs = ccrs.epsg(32635)
gv.Image(da, crs=crs) * tiles

```
</details>
  

An example is if I use CRS-32635. Which haves bounds from `(x1, y1, x2, y2) = (166021.44, 0.00, 534994.66, 9329005.18)`, which means that `y1` should always be able to plot if it is above 0. It will, with the current implementation, raise an error even though it shouldn't:

https://user-images.githubusercontent.com/19758978/181490725-fdc7bfc4-d0e4-4636-be19-91db787b42ac.mp4

This is because the `srj_proj.threshold` is 6680 and therefore ignores the first part of the data when doing the conversion.
https://github.com/holoviz/geoviews/blob/a315de762e1a3300877e210aac91798edc3ced29/geoviews/util.py#L88-L90

By setting `buffer(0)`, we get the desired result, but when we have data just outside the bounds, we get a thin line. I think this is somewhat related to the padding of the plot, but I'm not entirely sure. 

https://user-images.githubusercontent.com/19758978/181492466-b02d2856-8565-4b01-b35a-0205253a8ebd.mp4

If I add a keyword argument to `project_extents` of  `threshold=0` and then change the code above to `eroded_boundary = boundary_poly.buffer(-threshold)` and calculate the threshold before passing it to the function, here:

https://github.com/holoviz/geoviews/blob/a315de762e1a3300877e210aac91798edc3ced29/geoviews/plotting/plot.py#L72-L73

The way I calculate the threshold is very unelegant but is done with:

``` python
threshold = min(
    element._SheetCoordinateSystem__xstep,
    element._SheetCoordinateSystem__ystep,
    element.crs.threshold,
)
```

I get the desired error for the last example:

https://user-images.githubusercontent.com/19758978/181493623-02e1d4e5-005d-4bde-9f8a-7c69b885241c.mp4

geoviews/util.py
Hoxbro: This is actually what it already does. The IndexError is happening because there is no data after the conversion. I wanted to show a small example of it happening, but I could not get it to work without changing the code. See comment below.

Hoxbro:
This will now return this error message: 

``` python
da = xr.DataArray(np.arange(10000).reshape(100,100),
                  coords=dict(x=t + np.linspace(166021, 166021 + 3000, 100),
                              y=t + np.linspace(-3000, -15, 100)))
crs = ccrs.epsg(32635)
gv.Image(da, crs=crs)# * tiles
```

``` python-traceback

ValueError: Could not project data from 'WGS 84 / UTM zone 35N' projection to 'Mercator' projection. 
Ensure some data is inside the bounds +/- the threshold of the CRS. 
For 'WGS 84 / UTM zone 35N' this is (xmin, xmax, ymin, ymax) = (172701, 827299, 6680, 9322326).
```
---
The examples shown in the previous comment can be archived with the current implementation with:
``` python
import xarray as xr
import cartopy.crs as ccrs
import panel as pn
import geoviews as gv
import numpy as np

pn.extension()
gv.extension("bokeh", logo=False)

tiles = gv.tile_sources.Wikipedia().opts(width=1000, height=500, padding=0.1)
crs = ccrs.epsg(32635)
t = int(crs.threshold)


def plot(y1):
    da = xr.DataArray(np.arange(10000).reshape(100,100),
                      coords=dict(x=t + np.linspace(166021, 166021 + 3000, 100),
                                  y=t + np.linspace(y1, 3000, 100))
                     )
    return gv.Image(da, crs=crs) * tiles


w = pn.widgets.IntSlider(value=0, start=-2000, end=2000 , step=100)
pn.Column(w, gv.DynamicMap(pn.bind(plot, w)))
```

https://user-images.githubusercontent.com/19758978/183426898-17fe7824-074e-40e7-b01e-912add62ddb4.mp4

PR #601, opened 2022-12-05
Support python 3.11

Hoxbro:
Add support for python 3.11. For now, `numba` is not installed on python 3.11 until it is compatible. 

I have split the test into two files, one which runs pull request, push etc., using python versions 3.7, 3.9 and 3.11. And a weekly test that also runs the versions in between. 

I have also removed `actions/setup-python@v4` as it is not needed to run the test. 

At last I have added `echo $(python --version | tr ' [:upper:]' '=[:lower:]') > $CONDA_PREFIX/conda-meta/pinned` to the conda setup step to pin the python version.


maximlt:
> At last I have added echo $(python --version | tr ' [:upper:]' '=[:lower:]') > $CONDA_PREFIX/conda-meta/pinned to the conda setup step to pin the python version.

What does that do exactly?

I'll have a closer look tomorrow but at first sight I'm not thrilled by having to split up the test workflow into three files, specially if we have to do that across many repos, as there's just more change for things to diverge when they're more scattered. But I don't know maybe it's the only way!

Hoxbro:
> What does that do exactly?

It pins the python version, so it is not updated in the following `conda install`s [example](https://github.com/holoviz/geoviews/actions/runs/3623479448/jobs/6109387627#step:6:88).

1. `echo $(python --version | tr ' [:upper:]' '=[:lower:]')` outputs  the current python version installed with `major.minor.patch`, lowercasing it and replacing space with equal.  Example: `python=3.9.15`
2. `$CONDA_PREFIX` is the active environment location. 
3. `conda-meta/pinned` prevents packages listed from being updated [ref](https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-pkgs.html#preventing-packages-from-updating-pinning).

> ... not thrilled by having to split up the test workflow into three files

I'm not either, but I couldn't get it to work in a single file, and two (almost) identical files were a non-starter. I would _love_ to have it in one file.

maximlt:
> It pins the python version

Usually the issue I've seen with `doit develop_install` failing because of CPython being updated isn't that the version is updated but the *build* only. For instance what happened in one case I have in mind (maybe this is common, not sure) is that the first command (e.g. `conda create -n test-env python=3.x pyctdev`) installed Python built against OpenSSL 3.0.0. The next commands would try to pull dependencies that weren't built against OpenSSL 3.0.0 but OpenSSL 1.1.1, so conda downgrades CPython to a build of the same version but built against OpenSSL 1.1.1.

I think that unfortunately in this case this workaround wouldn't help :(

>  I would love to have it in one file.

I'll try to see if there's another way.

PR #603, opened 2022-12-09
Try to run test with shapely 2

Hoxbro:
The latest cartopy (v.0.21.0) is not compatible with shapely 2 but will be in the next release.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -19,7 +19,8 @@ jobs:
       matrix:
         os: ['ubuntu-latest', 'macos-latest', 'windows-latest']
         # Run on the full set on schedule, workflow_dispatch and push&tags events, otherwise on a subset.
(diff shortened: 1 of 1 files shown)


Hoxbro:
The only problem left with the upgrade is that this no longer works:
``` python
import geoviews as gv
import geoviews.feature as gf
import xarray as xr
from cartopy import crs

gv.extension('bokeh')
gf.land.opts(projection=crs.Geostationary())
```

Where these work `gf.land.opts(projection=crs.Geostationary(), scale="10m")` and `gf.land`. Another problem is that opts are saved on the `FeaturePlot` even if you instantiate it.

Hoxbro:
The last two commit (9eb60be and f12a1b3) fixes the last of the problems I mentioned in the previous post.

``` python
import geoviews as gv
import geoviews.feature as gf
from cartopy import crs

gv.extension('bokeh')
(gf.land().opts(projection=crs.Geostationary()) + gf.land()).opts(shared_axes=False)
```

Before:
Screenshot 2023-01-03 21 59 06

After:
Screenshot 2023-01-03 21 58 37

philippjfr (approved):
Thanks @Hoxbro! Pretty happy the changes look pretty minimal and all look good to me.

PR #612, opened 2023-01-09
Update docs assets and config

maximlt:
Mostly copy/pasting and adapting things from Panel, which is the reference for docs build. One day all this CSS/template will be upstreamed to nbsite.


Hoxbro:
Thank you for updating the documentation.

PR #625, opened 2023-04-16
Add Bokeh 3 support to GeoViews

Hoxbro:
This PR adds Bokeh 3 support to GeoViews. For this to work, Holoviews 1.16 is needed.


stonebig:
a geoviews pre-release compatible with bokeh-3.1 would be nice to have

Hoxbro:
There are still problems with annotations but I will fix that in another PR. Will merge this now so we can get a dev release out of Geoviews.

Issue #624, opened 2023-04-16
Gridded dataset notebook raise ValueError


Hoxbro:
Running the `examples/user_guide/Gridded_Datasets_I.ipynb` and `examples/user_guide/Gridded_Datasets_II.ipynb` with the latest param and Holoviews, we get the following error: `ValueError: kdims: list length must be between 2 and 2 (inclusive)`.

Related PRs: https://github.com/holoviz/holoviews/pull/5667 and https://github.com/holoviz/param/pull/605

![image](https://user-images.githubusercontent.com/19758978/232283130-43214c27-7b6f-47f1-b1d3-7c0f91ff3bd1.png)

``` python-traceback
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[4], line 4
      1 kdims = ['time', 'longitude', 'latitude']
      2 vdims = ['surface_temperature']
----> 4 xr_dataset = gv.Dataset(xr_ensemble, kdims=kdims, vdims=vdims)
      5 iris_dataset = gv.Dataset(iris_ensemble, kdims=kdims, vdims=vdims)

File ~/Repos/holoviz/geoviews/geoviews/element/geo.py:110, in _Element.__init__(self, data, kdims, vdims, **kwargs)
    108 elif isinstance(data, _Element):
    109     kwargs['crs'] = data.crs
--> 110 super(_Element, self).__init__(data, kdims=kdims, vdims=vdims, **kwargs)

File ~/Repos/holoviz/holoviews/holoviews/core/data/__init__.py:334, in Dataset.__init__(self, data, kdims, vdims, **kwargs)
    331 initialized = Interface.initialize(type(self), data, kdims, vdims,
    332                                    datatype=kwargs.get('datatype'))
    333 (data, self.interface, dims, extra_kws) = initialized
--> 334 super().__init__(data, **dict(kwargs, **dict(dims, **extra_kws)))
    335 self.interface.validate(self, validate_vdims)
    337 # Handle _pipeline property

File ~/Repos/holoviz/holoviews/holoviews/core/dimension.py:844, in Dimensioned.__init__(self, data, kdims, vdims, **params)
    841 if 'cdims' in params:
    842     params['cdims'] = {d if isinstance(d, Dimension) else Dimension(d): val
    843                        for d, val in params['cdims'].items()}
--> 844 super().__init__(data, **params)
    845 self.ndims = len(self.kdims)
[...]

PR #626, opened 2023-04-16
Param 2: allow `Dataset` to have any number of `kdims`

maximlt:
Fixes https://github.com/holoviz/geoviews/issues/624, in preparation of Param 2.

`kdims.bounds` was inherited from `_Element` with a value of `(2, 2)`, which appeared not to be valid for `Dataset`. Running the notebooks pointed out in the issue, I saw that `Dataset` could have more `kdims` than 2. I was also slightly surprised to see that it could have less than 2, I got an error where `kdims` was just `[Dimension('time')]`. I decided to let `kdims` accept any number of dimensions, i.e. `bounds=(0, None)`, as this anyway the default value in Param 1.

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -253,6 +253,7 @@ class Dataset(_Element, HvDataset):
     """
 
     kdims = param.List(default=[Dimension('Longitude'), Dimension('Latitude')],
+                       bounds=(0, None),
                        constant=True)
 
     group = param.String(default='Dataset')


Hoxbro:
Does it make sense to have none kdims? 

If it does, you can just merge this PR. If not, I suggest updating the bounds to `(1, None)`.

maximlt:
`kdims` is configured this way on the `Dimensioned` object in HoloViews:

https://github.com/holoviz/holoviews/blob/main/holoviews/core/dimension.py#L815

I don't know if it *makes sense* though.

PR #633, opened 2023-05-01
Annotation Bokeh 3 / Holoviews 1.16 support

Hoxbro:
- [x] Support Quad instead of Rect
- [x] Support Path and Polygon (actually, no problem with it as you have to select the object...)
- [x] ~Look into why data is not synced back for rectangle~ (holoviews issue: https://github.com/holoviz/holoviews/pull/5702)

### DIFF
--- a/geoviews/links.py
+++ b/geoviews/links.py
@@ -233,20 +233,18 @@ class RectanglesTableLinkCallback(HvRectanglesTableLinkCallback):
 
     source_code = """
     const projections = Bokeh.require("core/util/projections");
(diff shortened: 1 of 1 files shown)


djhoese:
What is the schedule for the final 1.10.0 release? The fixes in this PR are needed to fix compatibility with the version of Bokeh that conda/mamba are automatically pulling from conda-forge when creating new environments.

Hoxbro:
Planning to release it in the next couple of hours.

PR #645, opened 2023-05-31
Improve rioxarray support

### LINKED ISSUE #644: ValueError when Displaying GeoTIFFs with geoviews.util.from_xarray(rioxarray.open_rasterio(filename))
Hello GeoViews team! Thanks for taking the time to continuously maintain and improve this library. I'm having some difficulty trying to display TIFF files with the recent release (1.10.0), and I was hoping you could help me out.

#### ALL software version info
Geoviews 1.10.0 (Build pyhd8ed1ab_0, Channel conda-forge)
Bokeh 3.1.1 (Build pyhd8ed1ab_0,  Channel conda-forge)
Rioxarray 0.14.1 (Build pyhd8ed1ab_0, Channel conda-forge)
Python 3.10.11 (Build h4de0772_0_cpython, Channel conda-forge)
JupyterLab 4.0.0 (Build pyhd8ed1ab_1, Channel conda-forge)
OS: Windows 11 Home (Version 22H2, Build 22621.1702)
Browser: Google Chrome (Version 113.0.5672.127, Official Build, 64-bit)

#### Description of expected behavior and the observed behavior
I was able to display GeoTIFFs with `geoviews.util.load_tiff()` before it got deprecated.
expected_output
With the new release of GeoViews, I tried using the suggested `geoviews.util.from_xarray(rioxarray.open_rasterio(filename))` but I was not able to get the same Image element as before. I got a Dataset element instead, so I cannot apply any image options to it (see stack traceback below).

#### Complete, minimal, self-contained example code that reproduces the issue
```python
import geoviews as gv
import rioxarray as rxr
import cartopy.crs as ccrs
gv.extension("bokeh")

geotiff_img = gv.util.from_xarray(
    da = rxr.open_rasterio(filename = "https://github.com/mapbox/rasterio/raw/1.2.1/tests/data/RGB.byte.tif"),
    vdims = "Some Value",
    crs = ccrs.epsg(32618)
).opts(width = 500, height = 500, xaxis = None, yaxis = None)
geotiff_img
```

#### Stack traceback and/or browser JavaScript console output
```bash
ValueError                                Traceback (most recent call last)
Cell In[1], line 6
      3 import cartopy.crs as ccrs
      4 gv.extension("bokeh")
----> 6 geotiff_img = gv.util.from_xarray(
      7     da = rxr.open_rasterio(filename = "https://github.com/mapbox/rasterio/raw/1.2.1/tests/data/RGB.byte.tif"),
      8     vdims = "Some Value",
      9     crs = ccrs.epsg(32618)
     10 ).opts(width = 500, height = 500, xaxis = None, yaxis = None)
     11 geotiff_img

File ~\anaconda3\envs\test\lib\site-packages\holoviews\core\accessors.py:43, in AccessorPipelineMeta.pipelined.<locals>.pipelined_call(*args, **kwargs)
     40     inst._obj._in_method = True
     42 try:
---> 43     result = __call__(*args, **kwargs)
     45     if not in_method:
     46         init_op = factory.instance(
     47             output_type=type(inst),
     48             kwargs={'mode': getattr(inst, 'mode', None)},
     49         )

File ~\anaconda3\envs\test\lib\site-packages\holoviews\core\accessors.py:568, in Opts.__call__(self, *args, **kwargs)
    562         msg = ("Calling the .opts method with options broken down by options "
    563                "group (i.e. separate plot, style and norm groups) is deprecated. "
    564                "Use the .options method converting to the simplified format "
    565                "instead or use hv.opts.apply_groups for backward compatibility.")
    566         param.main.param.warning(msg)
--> 568 return self._dispatch_opts( *args, **kwargs)

File ~\anaconda3\envs\test\lib\site-packages\holoviews\core\accessors.py:572, in Opts._dispatch_opts(self, *args, **kwargs)
    570 def _dispatch_opts(self, *args, **kwargs):
    571     if self._mode is None:
--> 572         return self._base_opts(*args, **kwargs)
    573     elif self._mode == 'holomap':
    574         return self._holomap_opts(*args, **kwargs)

File ~\anaconda3\envs\test\lib\site-packages\holoviews\core\accessors.py:651, in Opts._base_opts(self, *args, **kwargs)
    648     return opts.apply_groups(self._obj, **dict(kwargs, **new_kwargs))
    650 kwargs['clone'] = False if clone is None else clone
--> 651 return self._obj.options(*new_args, **kwargs)

File ~\anaconda3\envs\test\lib\site-packages\holoviews\core\data\__init__.py:199, in PipelineMeta.pipelined.<locals>.pipelined_fn(*args, **kwargs)
    196     inst._in_method = True
    198 try:
--> 199     result = method_fn(*args, **kwargs)
    200     if PipelineMeta.disable:
    201         return result

File ~\anaconda3\envs\test\lib\site-packages\holoviews\core\data\__init__.py:1203, in Dataset.options(self, *args, **kwargs)
   1201 @wraps(Dimensioned.options)
   1202 def options(self, *args, **kwargs):
-> 1203     return super().options(*args, **kwargs)

File ~\anaconda3\envs\test\lib\site-packages\holoviews\core\dimension.py:1276, in Dimensioned.options(self, clone, *args, **kwargs)
   1274     expanded_backends = opts._expand_by_backend(options, backend)
   1275 else:
-> 1276     expanded_backends = [(backend, opts._expand_options(options, backend))]
   1278 obj = self
   1279 for backend, expanded in expanded_backends:

File ~\anaconda3\envs\test\lib\site-packages\holoviews\util\__init__.py:347, in opts._expand_options(cls, options, backend)
    345 objtype = objspec.split('.')[0]
    346 if objtype not in backend_options:
--> 347     raise ValueError(f'{objtype} type not found, could not apply options.')
    348 obj_options = backend_options[objtype]
    349 expanded[objspec] = {g: {} for g in obj_options.groups}

ValueError: Dataset type not found, could not apply options.
```

#### Screenshots or screencasts of the bug in action
I ran the example code in a Jupyter notebook and nothing except for the error gets outputted.

Hoxbro:
Fixes #644

Rioxarray `da.coords` contains`spatial_ref` with is not a dimension. This makes it go down the wrong code path. By only using `da.coords.dims`, we can make it go down the same code path as previously. 

I also aligned `process_crs` with the recent changes in hvplot's version of it in https://github.com/holoviz/hvplot/pull/1026. Though I added `+init` to the string comparison to avoid this warning:
```
/home/shh/miniconda3/envs/holoviz/lib/python3.11/site-packages/pyproj/crs/crs.py:141: FutureWarning: '+init=<authority>:<code>' syntax is deprecated. '<authority>:<code>' is the preferred initialization method. When making the change, be mindful of axis order changes: https://pyproj4.github.io/pyproj/stable/gotchas.html#axis-order-changes-in-proj-6
  in_crs_string = _prepare_from_proj_string(in_crs_string)
```

At last, I added simple crs detection for a rioxarray and updated `param.main.warning` to `warn` as it no longer exits.

These changes make it possible to run the original example like this:
image


geoviews/util.py
```diff
@@ -580,28 +579,37 @@ def process_crs(crs):
...
-    if isinstance(crs, str) and crs.lower().startswith('epsg'):
+    errors = []
+    if isinstance(crs, str):
+        starts = ['epsg:', '+init=epsg:']
+        for start in starts:
+            if crs.lower().startswith(start):
+                try:
+                    return ccrs.epsg(crs[len(start):].strip())
+                except Exception as e:
+                    errors.append(e)
```
ahuang11, line 597: If I'm understanding correctly, it's stripping the non-digits?

If so, I think this is more readable.

```suggestion
    if isinstance(crs, str):
        try:
            return ccrs.epsg("".join(c for c in crs if c.isdigit()))
        except Exception as e:
            errors.append(e)
```

Or if ccrs.epsg does accept non-digits:
`crs.lower().replace("epsg:", "").replace("+init=", "")`

geoviews/util.py
```diff
@@ -580,28 +579,37 @@ def process_crs(crs):
...
         try:
-            crs = ccrs.epsg(crs[5:].lstrip().rstrip())
-        except Exception:
-            raise ValueError("Could not parse EPSG code as CRS, must be of the format 'EPSG: {code}.'")
-    elif isinstance(crs, int):
-        crs = ccrs.epsg(crs)
-    elif isinstance(crs, str) or is_pyproj(crs):
+            return ccrs.epsg(crs)
+        except Exception as e:
+            crs = str(crs)
```
ahuang11, line 602:
```suggestion
            crs = str(crs)  # this will be converted to try again below
```

geoviews/util.py
```diff
@@ -677,13 +685,21 @@ def from_xarray(da, crs=None, apply_transform=False, nan_nodata=False, **kwargs)
...
+    elif hasattr(da, 'rio') and da.rio.crs is not None:
+        # rioxarray.open_rasterio
+        try:
+            kwargs['crs'] = process_crs(da.rio.crs.to_proj4())
+        except Exception:
+            warn(f'Could not decode projection from crs string {da.rio.crs}, '
+                  'defaulting to non-geographic element.')
 
-    coords = list(da.coords)
+    coords = list(da.coords.dims)
```
ahuang11, line 702:
```suggestion
    coords = list(da.dims)
```

geoviews/util.py
```diff
@@ -580,28 +579,37 @@ def process_crs(crs):
...
+    errors = []
+    if isinstance(crs, str):
+        starts = ['epsg:', '+init=epsg:']
+        for start in starts:
+            if crs.lower().startswith(start):
+                try:
+                    return ccrs.epsg(crs[len(start):].strip())
+                except Exception as e:
+                    errors.append(e)
+    if isinstance(crs, int):
```
ahuang11, line 598:
```suggestion
    elif isinstance(crs, int):
```

geoviews/util.py
```diff
@@ -580,28 +579,37 @@ def process_crs(crs):
...
-        except Exception:
-            raise ValueError("Could not parse EPSG code as CRS, must be of the format 'proj4: {proj4 string}.'")
-    elif not isinstance(crs, ccrs.CRS):
-        raise ValueError("Projection must be defined as a EPSG code, proj4 string, cartopy CRS or pyproj.Proj.")
-    return crs
+            return proj_to_cartopy(crs)
+        except Exception as e:
+            errors.append(e)
+    if isinstance(crs, ccrs.CRS):
+        return crs
```
ahuang11, line 610: I think this can be the first clause before the other ifs because it's cleanest.

```

if crs is None:
    return ccrs.PlateCarree()
elif isinstance(crs, ccrs.CRS):
    return crs

process
```

ahuang11 (commented):
Looks good to me; just left a few nits about readability!

geoviews/util.py
Hoxbro: I had the same idea but also didn't know if `epsg` did accept non-digits. 

Looking at https://en.wikipedia.org/wiki/EPSG_Geodetic_Parameter_Dataset it seems to be numbers between 1024 and 32767.

geoviews/util.py
Hoxbro: In general, I don't want to explain what the code does but why it does it. I feel like this suggestion falls into the latter category.

geoviews/util.py
Hoxbro: I'm not that familiar with xarray to know what the difference is between `da.coords.dims` and `da.dims`. Do you have a link to a quick explanation?

geoviews/util.py
Hoxbro: Agree

geoviews/util.py
Hoxbro: A small note is that it is faster to use list comprehension than a generator, see https://github.com/adamchainz/flake8-comprehensions/issues/156#issuecomment-1475149702.

geoviews/util.py
Hoxbro: I actually want to keep this as an if-statement. To avoid someone in the future changing the next `if` statement to an `elif`.

geoviews/util.py
ahuang11: Yeah I think it's only codes:
https://scitools.org.uk/cartopy/docs/latest/_modules/cartopy/crs.html#epsg

geoviews/util.py
ahuang11: I think they access the same thing; da.dims is just the shorthand (and I think more standard).

```
import xarray as xr
ds = xr.tutorial.open_dataset("air_temperature")
ds.coords["tau"] = ds.coords["time"] - ds.coords["time"][0]
print("DIMS", ds.dims)
print("COORDS", ds.coords)
print("DIMS", ds.coords.dims)
```

The ones prefixed with * are dimensions (dims)
image

geoviews/tests/test_util.py
```diff
@@ -0,0 +1,46 @@
...
+
+@pytest.mark.parametrize(
+    "raw_crs",
+    [
+        "+init=epsg:26911",
+        "4326",
+        4326,
+        "epsg:4326",
+        "EPSG: 4326",
+    ],
```
ahuang11, line 21:
```suggestion
        "EPSG: 4326",
        ccrs.PlateCarree(),
    ],
```

geoviews/util.py
Hoxbro: You are right. 

image

geoviews/util.py
Hoxbro: It is very negligible, but we should take what we can :upside_down_face:

ahuang11:
No problem, happy to review!

Issue #133, opened 2018-02-06
Add a Barbs element


philippjfr:
Since we have a VectorField element now it would be nice if we had a similar Barbs element. Cartopy and matplotlib already support drawing barbs and it shouldn't be too difficult to draw barbs using bokeh too.

Here's what barbs look like in cartopy:

![image](https://user-images.githubusercontent.com/1550771/35875078-20fbf7e0-0b66-11e8-9a7a-fc5054499bc6.png)

jbednar:
Sounds good.  For clarity, it would be helpful to paste a screenshot or link to another plotting program's barb glyphs to show what functionality would be expected, more concretely.

acrosby:
I am interested in visualizing points with barbs in addition to the vector field example above.

What might be involved in, or where to start, getting an implementation equivalent to MPL's (https://matplotlib.org/api/_as_gen/matplotlib.pyplot.barbs.html#matplotlib.pyplot.barbs) over to bokeh/geoviews?

Here's MPL's barb demo (https://matplotlib.org/gallery/images_contours_and_fields/barb_demo.html#sphx-glr-gallery-images-contours-and-fields-barb-demo-py):
![](https://matplotlib.org/_images/sphx_glr_barb_demo_001.png)

jbednar:
I've been wondering when someone would start asking for this! :-)

dwr-psandhu:
@jbednar Given mpl's implementation of barbs doesn't this issue belong to holoviews? I read the previous discussion that closed holoviz/holoviews#3901 based on cartopy's implementation.

PR #651, opened 2023-06-09
Add WindBarbs to Matplotlib Backend

ahuang11:
Still working on it. Also, I'm still on the fence whether this should reside in GeoViews vs HoloViews.

Closes https://github.com/holoviz/geoviews/issues/133

Wondering if this should follow VectorField conventions (magnitude and angle) instead of U, V

I'm leaning towards U and V because in my experience dealing with atmospheric science data, files are always separated by zonal and meridional winds (U and V).

The downside of not following VectorField conventions is that users wont be able to swap out gv.Barbs for gv.VectorField directly.

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -363,17 +363,19 @@ class WindBarbs(_Element, Selection2DExpr, HvGeometry):
         radians = np.pi / 2 - np.arctan2(-vs, -us)
 
         if isinstance(data, tuple):
-            transformed_data = (xs, ys, radians, uv_magnitudes)
+            reorganized_data = (xs, ys, radians, uv_magnitudes)
         else:
-            transformed_data = {}
+            # calculations on this data could mutate the original data
+            # here we do not do any calculations; we only store the data
+            reorganized_data = {}
             for kdim in kdims:
-                transformed_data[kdim] = data[kdim]
-            transformed_data["Angle"] = radians
-            transformed_data["Magnitude"] = uv_magnitudes
+                reorganized_data[kdim] = data[kdim]
+            reorganized_data["Angle"] = radians
+            reorganized_data["Magnitude"] = uv_magnitudes
             for vdim in vdims[2:]:
-                transformed_data[vdim] = data[vdim]
+                reorganized_data[vdim] = data[vdim]
             vdims = ["Angle", "Magnitude"] + vdims[2:]
-        return cls(transformed_data, kdims=kdims, vdims=vdims, **params)
+        return cls(reorganized_data, kdims=kdims, vdims=vdims, **params)
 
 
 class Image(_Element, HvImage):
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -351,6 +351,30 @@ class WindBarbs(_Element, Selection2DExpr, HvGeometry):
     vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
                                 Dimension('Magnitude')], bounds=(2, None))
 
+    @classmethod
+    def from_uv(cls, data, kdims=None, vdims=None, **params):
+        if isinstance(data, tuple):
+            xs, ys, us, vs = data
+        else:
+            us = data[vdims[0]]
+            vs = data[vdims[1]]
+
+        uv_magnitudes = np.hypot(us, vs)  # unscaled
+        radians = np.arctan2(-us, -vs)
+
+        if isinstance(data, tuple):
+            transformed_data = (xs, ys, radians, uv_magnitudes)
+        else:
+            transformed_data = {}
+            for kdim in kdims:
+                transformed_data[kdim] = data[kdim]
+            transformed_data["Angle"] = radians
+            transformed_data["Magnitude"] = uv_magnitudes
+            for vdim in vdims[2:]:
+                transformed_data[vdim] = data[vdim]
+            vdims = ["Angle", "Magnitude"] + vdims[2:]
+        return cls(transformed_data, kdims=kdims, vdims=vdims, **params)
+
 
 class Image(_Element, HvImage):
     """
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -360,7 +360,7 @@ class WindBarbs(_Element, Selection2DExpr, HvGeometry):
             vs = data[vdims[1]]
 
         uv_magnitudes = np.hypot(us, vs)  # unscaled
-        radians = np.arctan2(-us, -vs)
+        radians = np.pi / 2 - np.arctan2(-vs, -us)
 
         if isinstance(data, tuple):
             transformed_data = (xs, ys, radians, uv_magnitudes)
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -15,8 +15,10 @@ from holoviews.element import (
     QuadMesh as HvQuadMesh, Points as HvPoints,
     VectorField as HvVectorField, HexTiles as HvHexTiles,
     Labels as HvLabels, Rectangles as HvRectangles,
-    Segments as HvSegments
+    Segments as HvSegments, Geometry as HvGeometry,
 )
+from holoviews.element.selection import Selection2DExpr
+
 
 from shapely.geometry.base import BaseGeometry
 from shapely.geometry import (
@@ -335,6 +337,16 @@ class VectorField(_Element, HvVectorField):
     """
 
     group = param.String(default='VectorField', constant=True)
+
+    vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
+                                Dimension('Magnitude')], bounds=(1, None))
+
+
+class WindBarbs(_Element, Selection2DExpr, HvGeometry):
+    """
+    """
+    
+    group = param.String(default='WindBarbs', constant=True)
 
     vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
                                 Dimension('Magnitude')], bounds=(1, None))
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -348,8 +348,7 @@ class WindBarbs(_Element, Selection2DExpr, HvGeometry):
     
     group = param.String(default='WindBarbs', constant=True)
 
-    vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
-                                Dimension('Magnitude')], bounds=(1, None))
+    vdims = param.List(default=[Dimension('U'), Dimension('V')], bounds=(2, None))
 
 
 class Image(_Element, HvImage):
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -345,7 +345,7 @@ class VectorField(_Element, HvVectorField):
 class WindBarbs(_Element, Selection2DExpr, HvGeometry):
     """
     """
-    
+
     group = param.String(default='WindBarbs', constant=True)
 
     vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -348,7 +348,8 @@ class WindBarbs(_Element, Selection2DExpr, HvGeometry):
     
     group = param.String(default='WindBarbs', constant=True)
 
-    vdims = param.List(default=[Dimension('U'), Dimension('V')], bounds=(2, None))
+    vdims = param.List(default=[Dimension('Angle', cyclic=True, range=(0,2*np.pi)),
+                                Dimension('Magnitude')], bounds=(2, None))
 
 
 class Image(_Element, HvImage):
--- a/geoviews/plotting/mpl/chart.py
+++ b/geoviews/plotting/mpl/chart.py
@@ -22,62 +22,18 @@ class WindBarbsPlot(ColorbarPlot):
     to the minimum distance respectively.
     """
 
-    arrow_heads = param.Boolean(
-        default=True,
-        doc="""
-       Whether or not to draw arrow heads. If arrowheads are enabled,
-       they may be customized with the 'headlength' and
-       'headaxislength' style options.""",
-    )
+    rounding = param.Boolean(default=True, doc="""
+        Whether the vector magnitude should be rounded when allocating
+        barb components.""")
 
-    magnitude = param.ClassSelector(
-        class_=(str, dim),
-        doc="""
-        Dimension or dimension value transform that declares the magnitude
-        of each vector. Magnitude is expected to be scaled between 0-1,
-        by default the magnitudes are rescaled relative to the minimum
-        distance between vectors, this can be disabled with the
-        rescale_lengths option.""",
-    )
+    barb_increments = param.Dict(default=None, doc="""
+        A dictionary of increments specifying values to associate
+        with different parts of the barb.""")
+
+    flip_barb = param.Parameter(default=False, doc="""
+        Whether the lines and flags should point opposite to normal.""")
 
     padding = param.ClassSelector(default=0.05, class_=(int, float, tuple))
-
-    rescale_lengths = param.Boolean(
-        default=True,
-        doc="""
-       Whether the lengths will be rescaled to take into account the
-       smallest non-zero distance between two vectors.""",
-    )
-
-    # Deprecated parameters
-
-    color_index = param.ClassSelector(
-        default=None,
-        class_=(str, int),
-        allow_None=True,
-        doc="""
-        Deprecated in favor of dimension value transform on color option,
-        e.g. `color=dim('Magnitude')`.
-        """,
-    )
-
-    size_index = param.ClassSelector(
-        default=None,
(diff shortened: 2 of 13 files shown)


jbednar:
> whether this should reside in GeoViews vs HoloViews.

It would be odd to have something as specific as "WindBarbs" be in HoloViews; for it to fit semantically into HoloViews I think it would have to have a name and an interpretation that is not specific to wind. E.g. it could be something about representing additional dimensions in a VectorField, with an aside that suggests that it's common to do that in meteorology. If it's in HoloViews, following the VectorField conventions would be appropropriate.

Another reason to put it into GeoViews is that I don't believe HoloViews currently includes any custom Bokeh models or extensions, while implementing this for Bokeh will probably require those (which is ok for GeoViews).

jbednar:
>I'm leaning towards U and V because in my experience dealing with atmospheric science data, files are always separated by zonal and meridional winds (U and V).

Maybe there could be an alternative constructor (or constructor arguments) to convert incoming data from whichever of the two conventions is not the default. If data is natively in u,v it does make sense to keep it that way to avoid having to convert and store the data for most use cases.

ahuang11:
Good idea; I think we can do

`gv.WindBarbs.from_uv()`

ahuang11:
Okay, I just added a `from_uv_components` param

geoviews/plotting/mpl/chart.py
```diff
@@ -0,0 +1,156 @@
...
+    The smallest increment is a half line. There is only, of course, ever at most 1 half line.
+    If the magnitude is small and only needs a single half-line and no full lines or
+    triangles, the half-line is offset from the end of the barb so that it can be
+    easily distinguished from barbs with a single full line. The magnitude for the barb
+    shown above would nominally be 65, using the standard increments of 50, 10, and 5.
+    """
+
+    padding = param.ClassSelector(default=0.05, class_=(int, float, tuple))
+
+    from_uv_components = param.Boolean(default=False, doc="""
```
Hoxbro, line 27: I would add the data conversion to the `geoviews.element.geo.WindBarbs`. So it is possible in the future to use it for other backends.

My preference would be to have this functionality as a `classmethod`. So you could do something like WindBarbs.from_uv_componts(u, v)`.

geoviews/plotting/mpl/chart.py
```diff
@@ -0,0 +1,156 @@
...
+            magnitudes = element.dimension_values(mag_dim)
+
+        if self.invert_axes:
+            radians += 1.5 * np.pi
+
+        # Compute U, V to serialize to matplotlib
+        # Even if it's from U, V components, we still need to re-compute
+        # because operations may have been applied to magnitudes
+        if not self.from_uv_components or isinstance(mag_dim, dim):
+            us = magnitudes * np.cos(radians.flatten())
```
Hoxbro, line 103: As I read the code `radians` from `from_uv_components` is not in cartesian coordinates but in north-clockwise coordinates (or a better. Is this conversion still the same for this case?

geoviews/plotting/mpl/chart.py
```diff
@@ -0,0 +1,156 @@
...
+        if isinstance(mag_dim, str):
+            mag_dim = element.get_dimension(mag_dim)
+
+        if isinstance(mag_dim, dim):
+            magnitudes = mag_dim.apply(element, flat=True)
+        else:
+            magnitudes = element.dimension_values(mag_dim)
+
+        if self.invert_axes:
+            radians += 1.5 * np.pi
```
Hoxbro, line 97: This is just rotating the coordinates 270 degrees, correct? At the top of my head, I would think only 180 degrees are needed.

Hoxbro (commented):
Have added some comments. 

Some of them are trigonometry related. So stupid questions can occur. :slightly_smiling_face: 

I think the failing Mac test can be fixed by merge the main.

geoviews/plotting/mpl/chart.py
jbednar: Or just `from_uv`, as `uv` is a pretty strong indicator of `components`.

geoviews/plotting/mpl/chart.py
```diff
@@ -0,0 +1,156 @@
...
+    ]
+
+    _plot_methods = dict(single="barbs")
+
+    def _get_us_vs(self, element):
+        if self.from_uv_components:
+            us = element.dimension_values(2, flat=False) if len(element.data) else []
+            vs = element.dimension_values(3, flat=False) if len(element.data) else []
+            uv_magnitudes = np.hypot(us, vs)  # unscaled
+            radians = (np.pi / 2.0) - np.arctan2(us / uv_magnitudes, vs / uv_magnitudes)
```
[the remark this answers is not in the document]
Hoxbro: Can you explain this calculation to me? I'm not completely following the logic.

My understanding is `np.pi / 2 - ` is to convert from a cartesian coordinate to a compass coordinate system (or the other way around). Where the difference between those two is:
|   |  cartesian | compass |
| --- | ---- | --- | 
| rotation | counter-clockwise | clockwise |
| zero-axis | between 1. and 4. quadrants | between 1. and 2. quadrants | 




For the `arctan2`, is it correct that `us` should be seen as the `y` coordinate and `vs` as the `x` coordinate? I know that different languages swap the order so just want to be sure this is correct. [Numpys documentation](https://numpy.org/doc/stable/reference/generated/numpy.arctan2.html).

geoviews/plotting/mpl/chart.py
ahuang11: I believe this is correct because x / y axes are flipped. We don't want to just flip the direction.

image

ahuang11: As a sanity check: invert_axes just swaps the U and V args I think?
U, V
image


V, U
image

ahuang11:
To verify my implementation, I compared using metpy + matplotlib:
```python
# metpy barbs

import numpy as np
import metpy.calc as mpcalc
from metpy.units import units
import matplotlib.pyplot as plt

x = np.linspace(-1, 1, 4)
y = np.array([0, 1])
X, Y = np.meshgrid(x, y)
U, V = 10 * X, 0 * Y

mag = mpcalc.wind_speed(U * units("m/s"), V * units("m/s"))
angle = mpcalc.wind_direction(U * units("m/s"), V * units("m/s"))
us, vs = mpcalc.wind_components(mag, angle)

fig = plt.figure(figsize=(7, 15))
ax1 = plt.subplot(311)
ax1.barbs(X, Y, U, V)

ax2 = plt.subplot(312)
ax2.barbs(X, Y, us, vs)
ax2.set_title("UV metpy")

ax2 = plt.subplot(313)
ax2.barbs(Y, X, vs, us)
ax2.set_title("invert axes")
```

My implementation
```
import numpy as np
import pandas as pd
import geoviews as gv

gv.extension("matplotlib")

x = np.linspace(-1, 1, 4)
y = np.array([0, 1])
X, Y = np.meshgrid(x, y)
U, V = 10 * X, 0 * Y

angle = np.arctan2(-U, -V)
mag = np.hypot(U, V)

barbs = gv.WindBarbs((X, Y, angle, mag))
from_uv_barbs = gv.WindBarbs.from_uv((X, Y, U, V))
(barbs + from_uv_barbs + from_uv_barbs.clone().opts(invert_axes=True)).cols(1).opts(fig_inches=10)
```

Matplotlib
image

GeoViews
image

geoviews/plotting/mpl/chart.py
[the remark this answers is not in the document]
ahuang11: I think I fixed it. To disregard reasoning thru the trig, I just compared it against metpy + matplotlib which I believe should be good enough (but let me know if it isn't).
https://github.com/holoviz/geoviews/pull/651#issuecomment-1610239857

geoviews/plotting/mpl/chart.py
ahuang11: I think I fixed it. To disregard reasoning thru the trig, I just compared it against metpy + matplotlib which I believe should be good enough (but let me know if it isn't).
https://github.com/holoviz/geoviews/pull/651#issuecomment-1610239857

geoviews/plotting/mpl/chart.py
ahuang11: I made it rotate 90 degrees now.

geoviews/plotting/mpl/chart.py
ahuang11: Added `from_uv` classmethod.

geoviews/plotting/mpl/chart.py
```diff
@@ -0,0 +1,144 @@
...
+        else:
+            magnitudes = element.dimension_values(mag_dim)
+
+        if self.convention == "to":
+            radians -= np.pi
+
+        if self.invert_axes:
+            radians -= 0.5 * np.pi
+
+        us = -magnitudes * np.sin(radians.flatten())
```
ahuang11, line 91: Taken from https://github.com/Unidata/MetPy/blob/main/src/metpy/calc/basic.py#L151

geoviews/plotting/mpl/chart.py
```diff
@@ -0,0 +1,144 @@
...
+        radians = element.dimension_values(2) if len(element.data) else []
+
+        mag_dim = element.get_dimension(3)
+        if isinstance(mag_dim, dim):
+            magnitudes = mag_dim.apply(element, flat=True)
+        else:
+            magnitudes = element.dimension_values(mag_dim)
+
+        if self.convention == "to":
+            radians -= np.pi
```
ahuang11, line 86: Taken from https://github.com/Unidata/MetPy/blob/main/src/metpy/calc/basic.py#L98

geoviews/element/geo.py
```diff
@@ -340,6 +342,40 @@ class VectorField(_Element, HvVectorField):
...
+    @classmethod
+    def from_uv(cls, data, kdims=None, vdims=None, **params):
+        if isinstance(data, tuple):
+            xs, ys, us, vs = data
+        else:
+            us = data[vdims[0]]
+            vs = data[vdims[1]]
+
+        uv_magnitudes = np.hypot(us, vs)  # unscaled
+        radians = np.pi / 2 - np.arctan2(-vs, -us)
```
ahuang11, line 363: Taken from
https://github.com/Unidata/MetPy/blob/main/src/metpy/calc/basic.py#L92

geoviews/plotting/mpl/chart.py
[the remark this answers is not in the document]
ahuang11: Oh I got the arctan2 from https://geoviews.org/gallery/bokeh/orthographic_vectorfield.html#declare-data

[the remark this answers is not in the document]
Hoxbro: > which I believe should be good enough

I agree. 

> Oh, I got the arctan2 from https://geoviews.org/gallery/bokeh/orthographic_vectorfield.html#declare-data

It doesn't really matter for this example as the U and V are synthetic data, but let's update that example to use the same calculation as we use here.

geoviews/plotting/mpl/chart.py
Hoxbro: I forgot that `invert_axes` swap the axes and does not invert both axes independently, so this sound right.

geoviews/tests/plotting/mpl/test_chart.py
```diff
@@ -0,0 +1,197 @@
...
+
+gv.extension("matplotlib")
+mpl_renderer = Store.renderers["matplotlib"]
+
+
+class TestWindBarbsPlot(TestMPLPlot):
+    def test_windbarbs(self):
+        x = np.linspace(-1, 1, 4)
+        X, Y = np.meshgrid(x, x)
+        U, V = 10 * X, 0 * Y
```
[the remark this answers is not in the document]
ahuang11: I was testing visually to see if it matched expectation, but for tests I suppose I can make it whatever.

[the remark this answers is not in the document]
Hoxbro: It could be cool if you could make one of the tests parametrized with different angles.  

But not something which would block this PR.

geoviews/element/geo.py
```diff
@@ -340,6 +342,40 @@ class VectorField(_Element, HvVectorField):
...
+        if isinstance(data, tuple):
+            transformed_data = (xs, ys, radians, uv_magnitudes)
+        else:
+            transformed_data = {}
+            for kdim in kdims:
+                transformed_data[kdim] = data[kdim]
+            transformed_data["Angle"] = radians
+            transformed_data["Magnitude"] = uv_magnitudes
+            for vdim in vdims[2:]:
+                transformed_data[vdim] = data[vdim]
```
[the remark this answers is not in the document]
ahuang11: No I don't believe so; I create a new dict to hold it:
```
            transformed_data = {}
            for kdim in kdims:
                transformed_data[kdim] = data[kdim]
            transformed_data["Angle"] = radians
            transformed_data["Magnitude"] = uv_magnitudes
            for vdim in vdims[2:]:
                transformed_data[vdim] = data[vdim]
```

geoviews/tests/plotting/mpl/test_chart.py
```diff
@@ -0,0 +1,197 @@
...
+import geoviews as gv
+
+from geoviews.element import WindBarbs
+
+from holoviews.tests.plotting.matplotlib.test_plot import TestMPLPlot
+from holoviews.tests.plotting.utils import ParamLogStream
+
+from geoviews import Store
+
+gv.extension("matplotlib")
```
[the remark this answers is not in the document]
ahuang11: Thanks for catching that; I simply copied over from my notebook.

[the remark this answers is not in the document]
ahuang11: Okay, I don't think TestMPLPlot works without gv.extension because it's calling the HoloViews store instead of the GeoViews store.
```

        try:
            plotclass = Store.registry[cls.backend][element_type]
        except KeyError:
>           raise SkipRendering(f"No plotting class for {element_type.__name__} found.")
E           holoviews.core.options.SkipRendering: No plotting class for WindBarbs found.
```

geoviews/element/geo.py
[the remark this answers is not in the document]
Hoxbro: `transformed_data` is still holding a reference to `data`. An example could be if `data` is a DataFrame, then manipulating values of `transformed_data` will also manipulate `data`.

``` python
import numpy as np
import pandas as pd

data = pd.DataFrame({x: np.ones(10) for x in "xyuv"})
kdims, vdims = ["x", "y"], ["u", "v"]

transformed_data = {}
for kdim in kdims:
    transformed_data[kdim] = data[kdim]

transformed_data["y"] *= 10
data["y"]
```

image

I don't necessarily think we should copy the data, but at least be aware that changes to `transformed_data` could affect the original data.

geoviews/tests/plotting/mpl/test_chart.py
```diff
@@ -0,0 +1,197 @@
...
+
+        fig = gv.render(gv_barbs)
+        mpl_barbs = fig.axes[0].get_children()[0]
+        np.testing.assert_almost_equal(mpl_barbs.u.data, U.T.flatten())
+        np.testing.assert_almost_equal(mpl_barbs.v.data, V.T.flatten())
+
+    def test_windbarbs_dataset(self):
+        x = np.linspace(-1, 1, 4)
+        X, Y = np.meshgrid(x, x)
+        U, V = 10 * X, 0 * Y
```
Hoxbro, line 35: Let's change the other V too.

Hoxbro (commented):
Just in case you did not see [this comment](https://github.com/holoviz/geoviews/pull/651#discussion_r1252310808), I think we should update: https://geoviews.org/gallery/bokeh/orthographic_vectorfield.html#declare-data with the same calculations as used in here.

ahuang11:
Thanks for correcting me about the data mutation; that makes sense.

Here, we aren't actually doing any additional calculations on the original data; we simply reorganize the first two items to use angles and magnitudes. So instead of naming it transformed data, I renamed it to reorganized data and left a comment.

geoviews/tests/plotting/mpl/test_chart.py
```diff
@@ -0,0 +1,197 @@
...
+            },
+            coords={"x": x, "y": -x},
+        )
+
+        gv_barbs = gv.WindBarbs(ds, ["x", "y"], ["a", "m"])
+
+        fig = gv.render(gv_barbs)
+        mpl_barbs = fig.axes[0].get_children()[0]
+        np.testing.assert_almost_equal(mpl_barbs.u.data, U.T.flatten())
+        np.testing.assert_almost_equal(mpl_barbs.v.data, V.flatten())
```
Hoxbro, line 54:
```suggestion
        np.testing.assert_almost_equal(mpl_barbs.v.data, V.T.flatten())
```

I think this is the reason for the failing test.

ahuang11: Yeah thanks for catching that.

Hoxbro (approved):

ahuang11:
Woooo

philippjfr:
Sorry for not reviewing this earlier, and I might be missing something but does this implement proper projection support for the barbs?

ahuang11:
Probably not; I was just looking at https://github.com/holoviz/geoviews/pull/296

I can test it out soon.

PR #662, opened 2023-07-07
Support variety of data inputs for wind barbs

ahuang11:
Based on feedback from
https://github.com/holoviz/holoviews/pull/5797

### DIFF
--- a/geoviews/element/geo.py
+++ b/geoviews/element/geo.py
@@ -353,29 +353,30 @@ class WindBarbs(_Element, Selection2DExpr, HvGeometry):
 
     @classmethod
     def from_uv(cls, data, kdims=None, vdims=None, **params):
-        if isinstance(data, tuple):
-            xs, ys, us, vs = data
-        else:
-            us = data[vdims[0]]
-            vs = data[vdims[1]]
+        if kdims is None:
+            kdims = ['x', 'y']
+        if vdims is None:
+            vdims = ['u', 'v']
+        dataset = Dataset(data, kdims=kdims, vdims=vdims, **params)
+        us, vs = (dataset.dimension_values(i) for i in range(2, 4))
 
         uv_magnitudes = np.hypot(us, vs)  # unscaled
         radians = np.pi / 2 - np.arctan2(-vs, -us)
 
-        if isinstance(data, tuple):
-            reorganized_data = (xs, ys, radians, uv_magnitudes)
-        else:
-            # calculations on this data could mutate the original data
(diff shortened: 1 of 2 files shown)


geoviews/element/geo.py
```diff
@@ -353,29 +353,30 @@ class WindBarbs(_Element, Selection2DExpr, HvGeometry):
...
-            vs = data[vdims[1]]
+        if kdims is None:
+            kdims = ['x', 'y']
+        if vdims is None:
+            vdims = ['u', 'v']
+        dataset = Dataset(data, kdims=kdims, vdims=vdims, **params)
+        us, vs = (dataset.dimension_values(i) for i in range(2, 4))
 
         uv_magnitudes = np.hypot(us, vs)  # unscaled
         radians = np.pi / 2 - np.arctan2(-vs, -us)
```
Hoxbro, line 364: Let's add a comment that this is meteorological coordinates.

geoviews/tests/plotting/mpl/test_chart.py
```diff
@@ -54,7 +54,7 @@ def test_windbarbs_dataset(self):
         np.testing.assert_almost_equal(mpl_barbs.v.data, V.T.flatten())
 
     def test_windbarbs_from_uv(self):
-        x = np.linspace(-1, 1, 4)
+        x = np.linspace(-1, 1, 2)
```
Hoxbro, line 57:
```suggestion
        x = np.array([-1, 1])
```

Nit again.

Is there a reason for the change in the input array?

Hoxbro (commented):
Do you think we should add an extra test where `vdims` is `[hv.Dimension("u"), hv.Dimension("v")]`.

I don't know if this will give any problems but using `hv.Dataset` will flatten the array. I can see you already have flattened the result array because of this. 

``` python
import numpy as np
import holoviews as hv

x = np.array([-1, 1])
X, Y = np.meshgrid(x, x)
U, V = 10 * X, 2 * Y
U

ds = hv.Dataset((X, Y, U, V), kdims=["x", "y"], vdims=["u", "v"])
ds.dimension_values("u")
```
image

geoviews/tests/plotting/mpl/test_chart.py
ahuang11: I was testing why it was failing so I reduced the size so I can compare & contrast more easily; I changed it back.

Hoxbro (approved):
Let us keep this PR open to sync if more changes are added in https://github.com/holoviz/holoviews/pull/5797.

PR #664, opened 2023-07-18
Update to Bokeh 3.2 and update CI

Hoxbro:
I will wait with this merge until Bokeh 3.2.1 is released and use that for the Bokeh models. 

I'm also using this PR to get the CI up and running again for tests, docs, and builds.

Added the libmamba solver to the package build, and the time went from 1 hour and 11 minutes down to 18 minutes.


TheoMathurin:
Hi @Hoxbro 

Do you think there will be a micro-release including this when it's merged?

Issue #571, opened 2022-05-04
missing alpha in gv.load_tiff


oarcher:
it seems that the alpha channel is missing when loading a tiff file with `gv.load_tiff`

```python
!wget https://cerweb.ifremer.fr/oarcher/tmp/test_alpha.tiff
import geoviews as gv
gv.extension('bokeh')

gv.tile_sources.Wikipedia * gv.load_tiff('test_alpha.tiff')
```

![image](https://user-images.githubusercontent.com/4264660/166628047-f90697ac-d760-43e2-91b1-8fa5245e8ecd.png)

That's because in 
https://github.com/holoviz/geoviews/blob/8d7669f0b28dc022d7d0d90db4c342f405fc4476/geoviews/util.py#L731

`el.vdims` is `[Dimension('R'), Dimension('G'), Dimension('B')]`  (`A` dimension is missing)

A partial workaround is to replace with

```python
vdims = list('RGBA')[:bands]
```

![image](https://user-images.githubusercontent.com/4264660/166628772-b6be8040-d5f3-4248-8d20-209ee6f0526f.png)

But perhaps that `gv.RGB.vdims` should also return `Dimension('A')` ?

geoviews 1.9.5.post4+g8d7669f.dirty installed from master branch.

PR #671, opened 2023-09-14
Add alpha dimension

ahuang11:
Fixes https://github.com/holoviz/geoviews/issues/571
```python
import rioxarray
import geoviews as gv
gv.extension('bokeh')

ds = rioxarray.open_rasterio("test_alpha.tiff")
gv.util.from_xarray(ds, nan_nodata=True)
```
image

This is the most direct way to fix it, but there is also `alpha_dimension` here https://github.com/holoviz/holoviews/blob/main/holoviews/element/raster.py#L620-L626, which currently does not support xarray Dataset.


philippjfr:
Seems like there's something missing here, this only adds the test.

PR #685, opened 2023-10-31
Change default zoom_level on Matplotlib's WMTS 

ahuang11:
If users ran `gvts.OSM()` on a global extent with the original zoom=8, it'll fetch many many tiles, and probably gets the user rate limited.

I tried to use the `utils.zoom_level` to auto determine zoom, but I couldn't find a way to extract the bounds from WMTS.

```python
import geoviews as gv
from geoviews import opts, tile_sources as gvts

gv.extension('matplotlib')

gvts.OSM()
```

image

### DIFF
--- a/geoviews/util.py
+++ b/geoviews/util.py
@@ -745,8 +745,6 @@ def from_xarray(da, crs=None, apply_transform=False, nan_nodata=False, **kwargs)
         from .element.geo import RGB, HvRGB
         el = RGB if 'crs' in kwargs else HvRGB
         vdims = el.vdims[:bands]
-        if bands == 4:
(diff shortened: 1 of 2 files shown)


geoviews/util.py
```diff
@@ -745,6 +745,8 @@ def from_xarray(da, crs=None, apply_transform=False, nan_nodata=False, **kwargs)
         from .element.geo import RGB, HvRGB
         el = RGB if 'crs' in kwargs else HvRGB
         vdims = el.vdims[:bands]
+        if bands == 4:
+            vdims.append("A")
```
ahuang11, line 749:
```suggestion
```

PR #688, opened 2023-11-21
minor fix for #316

### LINKED ISSUE #316: Minimum zoom level
When a map is empty of elements, the extents are defaulted to extremely small (~10e-6). This causes any underlying tile service to error. If the user attempts to zoom back out, it can cause some strange windowing behavior inside the map. Being able to set a minimum zoom level would prevent this from happening.

dwr-psandhu:
fix #316


Hoxbro:
Thank you for correcting this :+1:

PR #709, opened 2024-02-19
Lazy load modules

Hoxbro:
Main:
image

This branch:
image

Ran with: `python -X importtime -c "import geoviews" 2> log.txt && tuna log.txt`


philippjfr:
Nice!

maximlt (approved):
Importing `annotate` dynamically means that the following top-level statements will no longer be executed when GeoViews is simply imported. It seems to me that it's fine since `annotate` is usually imported by users when they want to annotate geo data.

```python
Annotator._tools = [CheckpointTool, RestoreTool, ClearTool]
Annotator.table_transforms.append(project.instance(projection=ccrs.PlateCarree()))
PointAnnotator._link_type = get_point_table_link
RectangleAnnotator._link_type = get_rectangles_table_link
PathAnnotator._vertex_table_link = get_vertex_table_link
PolyAnnotator._vertex_table_link = get_vertex_table_link
Annotator._extra_opts['hooks'] = [initialize_tools]
annotate._annotator_types[Path] = PathBreakingAnnotator
```

PR #711, opened 2024-03-01
Doc: build the site without pulling the dev dependencies

maximlt:
I think the latest release of GeoViews partially broke the website, by pulling the latest dev version of Panel but not the latest version of Bokeh. As a consequence, Bokeh plots aren't displayed.

image
image

This PR builds the site using the generic install action and only using `conda-forge`. I think generally we don't want to pull pre-releases when building the websites. We sometimes need to add `pyviz/label/dev` to get the latest version of `nbsite`. By the way, I may need to either make a final release of nbsite or pull its latest dev version for doc build to succeed here, I'll see after manually running it for Github's UI.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -40,7 +40,7 @@ jobs:
           name: Documentation
           python-version: "3.10"
           channel-priority: strict
-          channels: conda-forge,nodefaults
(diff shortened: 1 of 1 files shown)


maximlt:
The *main* website has been re-deployed from this un-merged branch, it's fine again:

image

PR #713, opened 2024-03-20
Update build CI

Hoxbro:
Update build CI so it does not depend on pyctdev, split the build and publish up. 

image


scripts/conda/recipe-core/meta.yaml
```diff
@@ -1,15 +1,15 @@
...
+  version: {{ VERSION }}
 
 source:
-  path: ../..
+  url: ../../../dist/{{ sdata['name'] }}-{{ VERSION }}-py3-none-any.whl
 
 build:
   noarch: python
-  script: python setup.py install --single-version-externally-managed --record=record.txt
+  script: {{ PYTHON }} -m pip install -vv {{ sdata['name'] }}-{{ VERSION }}-py3-none-any.whl
```
maximlt, line 12: I'm curious, is that the standard approach for building a conda package, i.e. first run `python -m build ...` and then `conda build ...`, with the `script` set up to point to the wheel?

scripts/conda/recipe-recommended/meta.yaml
```diff
@@ -1,14 +1,15 @@
...
+  name: {{ sdata['name'] }}
+  version: {{ VERSION }}
 
 source:
-  path: ../..
+  url: ../../../dist/{{ sdata['name'] }}-{{ VERSION }}-py3-none-any.whl
 
 build:
   noarch: python
+  script: {{ PYTHON }} -m pip install -vv {{ sdata['name'] }}-{{ VERSION }}-py3-none-any.whl
```
Hoxbro, line 12:
```suggestion
```

scripts/conda/recipe-core/meta.yaml
Hoxbro: From a purist standpoint, it is not the correct way of doing it. You won't see it used on defaults or conda-forge as they want to build it from the bottom. 

From a maintainer standpoint, in my mind, it is so much easier to use the wheel and keep them separated. If a wheel can be built, then creating the conda package will not have a package-built problem. There could still be other problems with the `conda-build`, but that's at least one less hurdle to debug when things go wrong.

maximlt: Ok I see thanks. It's what Bokeh does too apparently, except they run it with `--no-deps`. Not sure what's the best practice there.

https://github.com/bokeh/bokeh/blob/69c084e692d79632c97ea1cfc5ceb10d0217e6f4/conda/recipe/meta.yaml#L13

Hoxbro: I don't have any objections to adding  `--no-deps`. You are welcome to add it.

PR #718, opened 2024-04-05
Add changelog

Hoxbro:
dask_expr import error spotted here: https://holoviz-dev.github.io/geoviews/user_guide/Resampling_Grids.html


geoviews/__init__.py
```diff
@@ -55,7 +55,7 @@ def __getattr__(attr):
     raise AttributeError(f"module {__name__} has no attribute {attr!r}")
 
 __all__ = [k for k in locals() if not k.startswith('_')]
-__all__ += ['annotate', 'project', 'operation']
+__all__ += ['annotate', 'project', 'operation', '__version__']
```
maximlt, line 58: Good catch!

Issue #673, opened 2023-09-18
geo=True and rasterize=True makes the plot disappear


ahuang11:
```python
import xarray as xr
import panel as pn
import holoviews as hv
from holoviews.operation.datashader import rasterize
import geoviews as gv
import cartopy.crs as ccrs

hv.renderer("bokeh").webgl = False
import hvplot.xarray
ds = xr.tutorial.open_dataset("air_temperature").isel(time=0)

rasterize((gv.Image(ds, ["lon", "lat"], ["air"]))) * gv.feature.coastline()
```
![image](https://github.com/holoviz/geoviews/assets/15331990/81280b20-beea-4ea6-841f-28fd8a3f64ac)

Data disappears on zoom.

ahuang11:
What about including datashader?

Hoxbro:
I used Datashader 1.15.2.

I did not expect Datashader to be a problem as it doesn't have any knowledge of geo-related information, and just get coordinates and data passed through it.

Did a test with 0.14.4, which reproduced the same results.

ahuang11:
I'm quite surprised that no one has commented about this yet besides me haha. I should probably dive deeper into this soon...

itsgifnotjiff:
I believe i had posted something about it a year or two ago but no one was interested so I deleted the post :) glad to see it ressurface.

ahuang11:
Workaround seems to work
```python
import xarray as xr
import panel as pn
import holoviews as hv
from holoviews.operation.datashader import rasterize
import geoviews as gv
import cartopy.crs as ccrs

hv.renderer("bokeh").webgl = False
import hvplot.xarray
ds = xr.tutorial.open_dataset("air_temperature").isel(time=0)
# pn.serve(ds.hvplot("lon", "lat", rasterize=True, crs=ccrs.PlateCarree(), projection=ccrs.PlateCarree()))

pn.serve(rasterize(gv.project(gv.Image(ds, ["lon", "lat"], ["air"]))))
```

ahuang11:
There are two separate issues here.

1. Bounciness with aspect(?) seen in hvplot:
```python
ds.hvplot("lon", "lat", geo=True, rasterize=True, coastline=True, project=True)
```

https://github.com/holoviz/geoviews/assets/15331990/72721652-be9e-41e7-b1ea-156a945e5665

No problem in geoviews


```python
import xarray as xr
import panel as pn
import holoviews as hv
from holoviews.operation.datashader import rasterize
[...]

Issue #593, opened 2022-10-11
Issue with resample_geometry + regridded gv.Image/gv.RGB with non-lat/lon bounds


TheoMathurin:
#### ALL software version info
GeoViews 1.9.5, HoloViews 1.15.1 on Linux

The code provided below is executed in a notebook (jupyter_client 7.0.6, jupyter_core 4.8.1, jupyter_server 1.11.1, notebook                  6.4.4) but I'm getting the same bug in a server (Panel 0.14.0).

This is not a regression as I've experienced this issue for more than a year with various versions.

#### Description of expected behavior and the observed behavior
When working in geostationary projection (or orthographic and presumably others) with Image or RGB elements, it is preferable to provide bounds calculated in the projection coordinates instead of lat/lon. This is because the corners of the element may not fit within the globe and therefore may not correspond to a valid (lat, lon) pair:

![image](https://user-images.githubusercontent.com/49269671/195099145-51049b43-fbd0-42e2-8ef3-2012a94f99ef.png)

When the image is large, regridding using `hd.regrid` allows to drastically reduce rendering time. Likewise, using `resample_geometry` on Feature objects (coastline typically) also helps.

However you can't use both at the same time. If you do, the plot does render, but the moment you pan or zoom, one of the element disappears (geometry or image, depending on their order in the overlay). The same thing happens if you use hv.Image/hv.RGB instead of their gv counterparts.

So in this case you have to choose between image regridding and feature resampling. However the bug does not arise if you provide bounds in lat/lon.

#### Complete, minimal, self-contained example code that reproduces the issue

```python
import numpy as np
import geoviews as gv
import holoviews.operation.datashader as hd
import cartopy.crs as ccrs

gv.extension('bokeh')

coastline = gv.feature.coastline.geoms('10m').opts(color='black')
resampled_coastline = gv.operation.resample_geometry(coastline)

projection = ccrs.Orthographic()
opts = dict(projection=projection, width=400, height=400)

ls = np.linspace(0, 10, 2000)
xx, yy = np.meshgrid(ls, ls)

# In separate cells:
latlon_bounds = (-20, -10, 20, 10)
latlon_img = gv.Image(np.sin(2*xx)*np.cos(yy), bounds=latlon_bounds)
regrid_latlon_img = hd.regrid(latlon_img)
(regrid_latlon_img*resampled_coastline).opts(**opts) # Works fine

# and
geo_bounds = (-2000000, -1000000, 2000000, 1000000)
[...]

PR #722, opened 2024-04-12
Fix x ranges upon projection

ahuang11:
Data no longer disappears upon zoom as described in https://github.com/holoviz/geoviews/issues/673#issuecomment-2050674997.

Closes https://github.com/holoviz/geoviews/issues/673

https://github.com/holoviz/geoviews/assets/15331990/9da07622-f4a6-4d63-bdce-abc27b8962b9

```python
import xarray as xr
import panel as pn
import holoviews as hv
from holoviews.operation.datashader import rasterize
import geoviews as gv
import cartopy.crs as ccrs

hv.renderer("bokeh").webgl = False
import hvplot.xarray
ds = xr.tutorial.open_dataset("air_temperature").isel(time=0)

rasterize((gv.Image(ds, ["lon", "lat"], ["air"]))) * gv.feature.coastline()
```

### DIFF
--- a/geoviews/plotting/bokeh/callbacks.py
+++ b/geoviews/plotting/bokeh/callbacks.py
@@ -122,8 +122,7 @@ class GeoRangeXYCallback(RangeXYCallback):
 
     def _process_msg(self, msg):
         msg = super()._process_msg(msg)
-        ranges = project_ranges(self, msg, ('x_range', 'y_range'))
-        return ranges
+        return project_ranges(self, msg, ('x_range', 'y_range'))
 
 class GeoRangeXCallback(RangeXCallback):
 
--- a/geoviews/plotting/bokeh/callbacks.py
+++ b/geoviews/plotting/bokeh/callbacks.py
@@ -64,6 +64,10 @@ def project_ranges(cb, msg, attributes):
     extents = x0, y0, x1, y1
     x0, y0, x1, y1 = project_extents(extents, plot.projection,
                                      plot.current_frame.crs)
+    if plot._unwrap_lons and -180 <= x0 < 0 or -180 <= x1 < 0:
+        x0, x1 = x0 + 360, x1 + 360
+        if x0 > x1:
+            x0, x1 = x1, x0
     coords = {'x_range': (x0, x1), 'y_range': (y0, y1)}
     return {k: v for k, v in coords.items() if k in attributes}
 
@@ -118,7 +122,8 @@ class GeoRangeXYCallback(RangeXYCallback):
 
     def _process_msg(self, msg):
         msg = super()._process_msg(msg)
-        return project_ranges(self, msg, ('x_range', 'y_range'))
+        ranges = project_ranges(self, msg, ('x_range', 'y_range'))
+        return ranges
 
 class GeoRangeXCallback(RangeXCallback):
 
--- a/geoviews/plotting/bokeh/plot.py
+++ b/geoviews/plotting/bokeh/plot.py
@@ -71,9 +71,6 @@ class GeoPlot(ProjectionPlot, ElementPlot):
                 )
 
         self._unwrap_lons = False
-        if isinstance(self.geographic, _CylindricalProjection):
-            x1, x2 = element.range(0)
-            self._unwrap_lons = 0 <= x1 <= 360 and 0 <= x2 <= 360
 
     def _axis_properties(self, axis, key, plot, dimension=None,
                          ax_mapping=None):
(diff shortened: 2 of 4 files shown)


geoviews/plotting/bokeh/callbacks.py
```diff
@@ -118,7 +122,8 @@ class GeoRangeXYCallback(RangeXYCallback):
 
     def _process_msg(self, msg):
         msg = super()._process_msg(msg)
-        return project_ranges(self, msg, ('x_range', 'y_range'))
+        ranges = project_ranges(self, msg, ('x_range', 'y_range'))
+        return ranges
```
jbednar, line 126: Doesn't seem to add anything?

jbednar (commented):
Thanks for addressing this. I can't quite decide if it's the right fix, though; would be good to have e.g. @philippjfr think about it.

geoviews/plotting/bokeh/callbacks.py
ahuang11: Oh yeah, I can revert that; it was mainly to printing it out during debugging.

ahuang11, line 126:
```suggestion
        return project_ranges(self, msg, ('x_range', 'y_range'))
```

ahuang11:
My other two thoughts about approaches to fixing this is:

1. require -180 to 180 for PlateCarree, else raise an error
2. same as 1, but raise a warning instead
3. if 0-360 is detected, automatically convert the data (potentially costly to load data)

This PR's fix only updates two integers, which is basically "free"

geoviews/plotting/bokeh/plot.py
```diff
@@ -70,6 +70,11 @@ def __init__(self, element, **params):
                     'and disable the show_grid option.'
                 )
 
+        self._unwrap_lons = False
+        if isinstance(self.geographic, _CylindricalProjection):
+            x1, x2 = element.range(0)
```
philippjfr, line 75: I think element here is likely misnamed, we probably have to move this into some other method, e.g. `initialize_plot` and `update_frame`.

TheoMathurin:
Maybe this can possibly also fix #593?

ahuang11:
Maybe! If it uses lat/lon, then there's a chance.

geoviews/plotting/bokeh/plot.py
```diff
@@ -113,8 +120,15 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
                                      overlaid=True, renderer=self.renderer)
             shapeplot.geographic = False
             shapeplot.initialize_plot(plot=fig)
+        nonempty = [(k, el) for k, el in self.hmap.data.items() if el]
+        self._set_unwrap_lons(nonempty[-1][1])
         return fig
 
+    def update_frame(self, key, ranges=None, element=None):
+        if element:
```
philippjfr, line 128:
```suggestion
        if element is not None:
```

geoviews/plotting/bokeh/plot.py
```diff
@@ -113,8 +120,15 @@ def initialize_plot(self, ranges=None, plot=None, plots=None, source=None):
                                      overlaid=True, renderer=self.renderer)
             shapeplot.geographic = False
             shapeplot.initialize_plot(plot=fig)
+        nonempty = [(k, el) for k, el in self.hmap.data.items() if el]
+        self._set_unwrap_lons(nonempty[-1][1])
```
philippjfr, line 124:
```suggestion
        self._set_unwrap_lons(self.current_frame)
```

ahuang11:
The changes work when it's not overlaid with coastline.

```python
import xarray as xr
import panel as pn
import holoviews as hv
from holoviews.operation.datashader import rasterize
import geoviews as gv
import cartopy.crs as ccrs

hv.extension("bokeh")
import hvplot.xarray
ds = xr.tutorial.open_dataset("air_temperature").isel(time=0)

rasterize(gv.Image(ds, ["lon", "lat"], ["air"]))
```

However, once coastline is overlaid, it produces a blank image.
image

geoviews/plotting/bokeh/callbacks.py
```diff
@@ -64,6 +64,10 @@ def project_ranges(cb, msg, attributes):
     extents = x0, y0, x1, y1
     x0, y0, x1, y1 = project_extents(extents, plot.projection,
                                      plot.current_frame.crs)
+    if plot._unwrap_lons and -180 <= x0 < 0 or -180 <= x1 < 0:
```
philippjfr, line 67: This should be:

```
if plot._unwrap_lons and (-180 <= x0 < 0 or -180 <= x1 < 0):
```

Issue #721, opened 2024-04-11
geoviews throws TypeError: '<=' not supported between  instances of 


dwr-psandhu:
# Versions of relevant libraries (python 3.11 env)
bokeh                     3.3.4              pyhd8ed1ab_0    conda-forge
geopandas                 0.14.3             pyhd8ed1ab_0    conda-forge
geopandas-base            0.14.3             pyha770c72_0    conda-forge
geoviews                  1.11.1             pyhd8ed1ab_0    conda-forge
geoviews-core             1.11.1             pyha770c72_0    conda-forge
holoviews                 1.18.3             pyhd8ed1ab_0    conda-forge
# Unncessary calculation of range on columns
geopandas array with many columns some of which of items that are not comparable (i.e. different types or Nans)

# Example

```
# %%
import geopandas as gpd
import pandas as pd
import holoviews as hv
import geoviews as gv

hv.extension("bokeh")
import cartopy.crs as ccrs

# %%
# create a dataframe with some x,y coordinates and a couple of columns
df = pd.DataFrame(
    {
        "x": [1, 2, 3, 4, 5],
        "y": [1, 2, 3, 4, 5],
        "value1": [10, 20, 30, 40, 50],
        "value2": [5, '4', 3, 2, 1],
    }
)
# emulate UTM zone 10n coordinates
df["x"] = df["x"].astype(float) + 625000
df["y"] = df["y"].astype(float) + 4180000
# %%
gdf = gpd.GeoDataFrame(df, geometry=gpd.points_from_xy(df.x, df.y))

# %%
gv.Points(gdf)
# %%
```

#### Stack traceback and/or browser JavaScript console output
```
{
	"name": "TypeError",
	"message": "'<=' not supported between instances of 'int' and 'str'",
	"stack": "---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
File c:\\Users\\psandhu\\AppData\\Local\\miniconda3\\envs\\pydelmod\\Lib\\site-packages\\IPython\\core\\formatters.py:977, in MimeBundleFormatter.__call__(self, obj, include, exclude)
    974     method = get_real_method(obj, self.print_method)
    976     if method is not None:
--> 977         return method(include=include, exclude=exclude)
    978     return None
    979 else:

[...]

PR #723, opened 2024-04-14
Return nans for range if column can't find min and max

hoxbro:
Fixes https://github.com/holoviz/geoviews/issues/721


geoviews/data/geopandas.py
```diff
@@ -284,7 +284,10 @@ def range(cls, dataset, dim):
                 return bounds.miny.min(), bounds.maxy.max()
         else:
             vals = dataset.data[dim.name]
-            return vals.min(), vals.max()
+            try:
+                return vals.min(), vals.max()
+            except Exception:
```
Hoxbro, line 289:
```suggestion
            except TypeError:
```

PR #763, opened 2024-11-01
compat: rasterio 1.4.2

hoxbro:
See: https://github.com/holoviz/geoviews/actions/runs/11631614282/job/32393057533

image

PR #764, opened 2024-11-04
perf: Postpone bokeh.models import

hoxbro:
`bokeh.models` is a heavy module to import, so we will postpone it until something else imports it.


hoxbro:
Failing test are fixed with https://github.com/holoviz/geoviews/pull/763

PR #789, opened 2025-05-14
Convert docstrings to Numpydoc style

### LINKED ISSUE #788: docs: Convert docstring to using numpy doc style
Convert all public API docstrings to use numpydoc format as done in [HoloViews#6506](https://github.com/holoviz/holoviews/pull/6506)

Azaya89:
fixes #788 

- [x] Convert all public API docstrings to use numpydoc format
- [x] Exclude the `tests/` dir
- [x] Configure ruff linting to use numpydoc style

### DIFF
--- a/geoviews/data/geopandas.py
+++ b/geoviews/data/geopandas.py
@@ -579,9 +579,9 @@ def to_geopandas(data, xdim, ydim, columns=None, geom='point'):
         apart from the geometry. Defaults to an empty list.
     geom : {'point', 'Line', 'Polygon'}, default='point'
         Specifies the geometry type to construct. Supports:
-        - 'point': Point or MultiPoint
+        - 'point' : Point or MultiPoint
         - 'Line' : LineString or MultiLineString
-        - 'Polygon': Polygon or MultiPolygon
+        - 'Polygon' : Polygon or MultiPolygon
 
     Returns
     -------
--- a/geoviews/data/geopandas.py
+++ b/geoviews/data/geopandas.py
@@ -564,22 +564,30 @@ def get_geom_type(geom):
 
 
 def to_geopandas(data, xdim, ydim, columns=None, geom='point'):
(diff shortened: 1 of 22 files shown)


geoviews/data/geopandas.py
```diff
@@ -562,14 +566,20 @@ def get_geom_type(geom):
 def to_geopandas(data, xdim, ydim, columns=None, geom='point'):
     """Converts list of dictionary format geometries to spatialpandas line geometries.
```
Azaya89, line 567: I'm not quite sure this docstring accurately describes what the `to_geopandas` function does.

geoviews/element/geo.py
```diff
@@ -311,20 +303,18 @@ class Dataset(_Element, HvDataset):
...
 
     def geom(self, union=False, projection=None):
-        """
-        Converts the Points to a shapely geometry.
+        """Converts the Points to a shapely geometry.
 
         Parameters
         ----------
-        union: boolean (default=False)
+        union: boolean, default=False
```
maximlt, line 317: There must be a space between the parameter name and the colon (https://numpydoc.readthedocs.io/en/latest/format.html#parameters) (I haven't checked the whole PR, just spotted this one).

```suggestion
        union : boolean, default=False
```

maximlt (commented):
Panel builds its API reference using the `numpydoc` extension instead of `sphinx.ext.napoleon` (https://github.com/holoviz/panel/blob/9321b1851deef57094681b4a9af07865290296d4/doc/conf.py#L89). I set up hvPlot to build its API reference the same way, as I found the website looked better this way (IIRC). If the same is done with GeoViews, it is possible the docstrings will have to be touched a little to fit what numpydoc's parser.

hoxbro:
Let's merge this as is and fix small mistakes in a follow-up PR, which enables the API on the documentation.

hoxbro (approved):

PR #792, opened 2025-06-16
Add mask_extrapolated option to project_image

philippjfr:
Image projection without mask_extrapolated causes weird artifacts in most cases so we enable it by default and expose it as an option.

Fixes issue mentioned in https://github.com/holoviz/hvplot/pull/1600

### DIFF
--- a/geoviews/operation/projection.py
+++ b/geoviews/operation/projection.py
@@ -369,7 +369,9 @@ class project_image(_project_operation):
     """Projects an geoviews Image to the specified projection,
     returning a regular HoloViews Image type. Works by
     regridding the data along projected bounds. Only supports
-    rectangular projections.
+    rectangular projections. If mask_extrapolated, will
+    additionally wrap longitudes from 0:360 to -180:180
+    if the source coordinate is cylindrical.
     """
 
     fast = param.Boolean(default=False, doc="""
@@ -420,8 +422,6 @@ class project_image(_project_operation):
             arr = img.dimension_values(vd, flat=False)
             if arr.size:
                 if self.p.mask_extrapolated:
-                    # Do not do additional computation if unnecessary
-                    # https://github.com/holoviz/geoviews/pull/792
                     src_extent = (
                         *wrap_cylindrical_projection_lons(
                             img.crs, src_extent[0], src_extent[1]


ahuang11:
Added the tests, and made the default = False to prevent breaking changes:

philippjfr:
I strongly vote to enable this by default and find a fix for the failure condition.

ahuang11:
>  find a fix for the failure condition.

The failure condition is just if the user doesn't first wrap their longitudes to -180 to 180.

We can do that calculation, very easily `xs_converted = ((xs + 180) % 360) - 180` and I think we have something that does that: https://github.com/holoviz/geoviews/blob/5092afed9e366767d442ed67dffd637e431e5ffc/geoviews/plotting/bokeh/plot.py#L105-L117
https://github.com/holoviz/geoviews/blob/5092afed9e366767d442ed67dffd637e431e5ffc/geoviews/util.py#L79

The thing is sorting the data I think which is computationally expensive.

Alternatively the easy solution is to check for 0 to 360 and raise if exceeds > 180.

philippjfr:
I don't have any particular worry about maintaining lazyness. `project_image` has to materialize the data anyway and usually projecting happens as a last step before plotting. I think we should apply the wrapping and enable the masking by default.

ahuang11:
```
import xarray as xr
import hvplot.xarray
import numpy as np
from geoviews import Image
from geoviews.operation import project_image
import cartopy.crs as ccrs

xs = np.linspace(200, 330, 5)
# xs = (xs + 180) % 360 - 180  # Wrap longitudes to [-180, 180]; no longer necessary
ys = np.linspace(-60, 60, 3)
zs = np.sin(np.linspace(0, 2 * np.pi, len(ys)))[:, None] * np.sin(np.linspace(0, 2 * np.pi, len(xs)))[None, :]
img = Image((xs, ys, zs), crs=ccrs.PlateCarree())

project_image(img, projection=ccrs.PlateCarree(), mask_extrapolated=True)
```

geoviews/operation/projection.py
```diff
@@ -386,6 +387,11 @@ class project_image(_project_operation):
         when applying project_image, backends that support linked streams
         update RangeXY streams on the inputs of the operation.""")
 
+    mask_extrapolated = param.Boolean(default=False, doc="""
```
hoxbro, line 390:
```suggestion
    mask_extrapolated = param.Boolean(default=True, doc="""
```

We want this on by default, so hvplot can get the benefit.

geoviews/util.py
```diff
@@ -39,6 +39,19 @@ def wrap_lons(lons, base, period):
...
 
 
+def wrap_cylindrical_projection_lons(src_proj, x1, x2, base=-180.0, period=360.0):
+    # Wrap longitudes
+    cx1, cx2 = src_proj.x_limits
+    if isinstance(src_proj, ccrs._CylindricalProjection):
+        lons = wrap_lons(np.linspace(x1, x2, 10000), base, period)
+        x1, x2 = lons.min(), lons.max()
+    else:
+        print('hello')
```
hoxbro, line 49: debug print statement.

hoxbro (changes requested):

hoxbro:
Can you fix the tests?

geoviews/operation/projection.py
```diff
@@ -413,8 +419,19 @@ def _process(self, img, key=None):
         for vd in img.vdims:
             arr = img.dimension_values(vd, flat=False)
             if arr.size:
-                projected, _ = warp_array(arr, proj, img.crs, (xn, yn),
-                                          src_extent, tgt_extent)
+                if self.p.mask_extrapolated:
+                    # Do not do additional computation if unnecessary
+                    # https://github.com/holoviz/geoviews/pull/792
```
hoxbro, line 424: I don't think this comment is relevant.

geoviews/tests/test_projection.py
```diff
@@ -76,3 +76,57 @@ def test_project_windbarbs(self):
...
+
+        # Create image with longitude data from 200 to 330 (outside standard range)
+        xs = np.linspace(200, 330, 5)
+        ys = np.linspace(-60, 60, 3)
+        zs = np.ones((3, 5))  # Fill with constant value
+        img = Image((xs, ys, zs), crs=ccrs.PlateCarree())
+
+        # Test with mask_extrapolated=True (default)
+        proj_op_masked = project_image.instance(projection=ccrs.PlateCarree())
+        proj_op_masked.mask_extrapolated = True
```
hoxbro, line 100:
```suggestion
```

If it is the default, we shouldn't set it.

geoviews/tests/test_projection.py
```diff
@@ -17,9 +17,9 @@ def test_image_latlon360_wrapping(self):
         proj = project(img, projection=ccrs.PlateCarree())
         zs = proj.dimension_values('z', flat=False)
         self.assertEqual(zs, np.array([
-            [-12960., -17280., -21600.,  -4320.,  -8640.],
+            [ -4320.,  -8640., -12960., -17280., -21600.],
```
hoxbro, line 20: Does this still look correct?

hoxbro (approved):

PR #813, opened 2025-10-23
fix: Missing borders

hoxbro:
Used the following MRE, and used AI to detect where the problem occurred. Still need to think more about the solution. 

```python
import cartopy.feature as cfeature
import holoviews as hv

import geoviews as gv

gv.extension("bokeh")

data = list(cfeature.BORDERS.with_scale("110m").geometries())

hv_len = len(hv.render(hv.Path(data)).renderers[0].data_source.data["xs"])
gv_len = len(hv.render(gv.Path(data)).renderers[0].data_source.data["xs"])

assert hv_len == 331, hv_len
assert gv_len == 331, gv_len
```

Notice how the borders between Germany and Poland are missing on the main branch.

This PR:

image

Before:

image

### DIFF
--- a/geoviews/operation/projection.py
+++ b/geoviews/operation/projection.py
@@ -114,7 +114,7 @@ class project_path(_project_operation):
             elif (not geom or isinstance(geom, GeometryCollection)):
                 continue
 
-            proj_geom = proj.project_geometry(geom, element.crs)
+            proj_geom = proj.project_geometry(geom, crs)
 
             # Attempt to fix geometry without being noisy about it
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -37,7 +37,7 @@ dependencies = [
     'panel >=1.0.0',
     'param >=1.9.3,<3.0',
     'pyproj',
-    'shapely',
+    'shapely >=1.8.0',
```
hoxbro: Was released in October 25th, 2021

image

PR #844, opened 2026-06-24
ci: Fix tests suite

hoxbro:
Fixes failing UI test.

### DIFF
--- a/examples/conftest.py
+++ b/examples/conftest.py
@@ -1,4 +1,15 @@
+import re
+from importlib.metadata import version
 from importlib.util import find_spec
+
+_re_no = re.compile(r"\d+")
+
+def get_version(package):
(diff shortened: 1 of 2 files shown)


examples/conftest.py
```diff
@@ -23,6 +34,9 @@
         "user_guide/Working_with_Bokeh.ipynb",
     ]
 
+if MPL_VERSION >= (3, 11, 0) and CARTOPY_VERSION <= (0, 25, 0):
+    # https://github.com/SciTools/cartopy/issues/2682
+    collect_ignore_glob += ["user_guide/Geometries.ipynb"]
```
hoxbro: Upstream problem and need a new cartopy release. Have tested the workaround in the issue and it works.

PR #874, opened 2026-09-12
chore: bump nbsite

hoxbro:
## Description

Because the fix in https://github.com/holoviz-dev/nbsite/pull/362

And GeoViews build currently fails because of that:

image

## Checklist

- [x] Tests added and are passing


hoxbro:
Did the trick:

image

PR #877, opened 2026-09-14
build: Work with Panel which supports Bokeh 3.10

hoxbro:
## Description

image


hoxbro:
Also added infrastructure to run build and docs with label

PR #884, opened 2026-09-18
compat: Cartopy 0.26.0

rajeeja:
## Problem

Cartopy 0.26 switched `PlateCarree` from `proj=eqc` to `proj=latlong`, which reports the central longitude as the prime meridian `pm` and no longer emits `lon_0`:

```python
>>> import cartopy.crs as ccrs
>>> ccrs.PlateCarree(central_longitude=30).proj4_params
{'ellps': 'WGS84', 'a': 6378137.0, 'proj': 'latlong', 'pm': 30}
```

`project_extents` (`geoviews/util.py:70,72`) and `project_path._process_element` (`geoviews/operation/projection.py:89`) both read `proj4_params["lon_0"]` directly. Under cartopy 0.26 any projected polygon or path plot from a PlateCarree source to another projection raises `KeyError: 'lon_0'` at render time, including the common case where the central longitude is 0, because the lookup happens before the `!= 0` comparison.

```python
import cartopy.crs as ccrs
import geoviews as gv
from geoviews.operation.projection import project_path

path = gv.Path([[(-10, -10), (10, 10)]], crs=ccrs.PlateCarree(central_longitude=30))
project_path(path, projection=ccrs.Robinson())
```

```
  File "geoviews/operation/projection.py", line 93, in _process_element
    and crs.proj4_params["lon_0"] != 0
        ~~~~~~~~~~~~~~~~^^^^^^^^^
KeyError: 'lon_0'
```

Found while moving UXarray off its `matplotlib<3.11` pin (UXARRAY/uxarray#1772). Cartopy 0.26 is what unblocks matplotlib 3.11 for downstreams, so this is the remaining blocker for anyone taking that upgrade.

## Fix

Adds a `central_longitude` helper to `geoviews/util.py` that prefers `lon_0`, falls back to `pm`, and returns 0 when neither is present, then routes both call sites through it. Behaviour on cartopy < 0.26 is unchanged, since `lon_0` is still preferred where it exists.

## Tests

`test_central_longitude` parametrizes both proj4 layouts against a stub, so both stay covered whichever cartopy is installed. `test_project_extents_offset_platecarree` and `test_project_path_from_offset_platecarree` exercise the two real call sites; both fail on `main` with `KeyError: 'lon_0'` under cartopy 0.26 and pass with this change.

`geoviews/tests/test_util.py` and `geoviews/tests/test_projection.py`: 26 passed, 1 skipped on cartopy 0.26.0, and the same on cartopy 0.25.0.


rajeeja:
Any chance of a new release once this lands? cartopy 0.26 breaks projected plots and downgrading isn't an option if you need matplotlib 3.11 too.

hoxbro:
I hope we can release in the next couple of weeks.

rajeeja:
> I hope we can release in the next couple of weeks.  

Thanks 🙏

